PR #54, opened 2015-02-20
rejiggle layout for additional clarity

DomT4:
Hopefully this is useful.

Closes https://github.com/Homebrew/homebrew/issues/36970

### DIFF
--- a/index_ko.html
+++ b/index_ko.html
@@ -4,21 +4,21 @@ lang: ko
 subtitle: OS X 용 패키지 관리자
 
 pagecontent:
-  what: Homebrew는 Apple에서 제공하지 않지만 <a href="https://github.com/Homebrew/homebrew/tree/master/Library/Formula" title="List of Homebrew packages">필요한 패키지</a>를 설치합니다.
-  how: Homebrew는 자신의 디렉토리에 패키지를 설치하고, <code>/usr/local</code>안에 그 파일을 심볼링 링크로 겁니다.
-  prefix: Homebrew는 prefix가 붙지 않는 파일은 설치 하지 않지만, 여러분이 원하는 위치 어느곳이든 설치할 수 있습니다.
-  createpackages: 자신의 Homebrew 패키지를 간편하게 만들 수 있습니다.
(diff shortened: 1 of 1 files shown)


DomT4:
[What it looks like](https://domt4.github.io/).

melodykramer:
oh yeah, that's beautiful.

_layouts/index.html
```diff
@@ -54,74 +54,88 @@
       <div id="informations">
         <ul>
           <li>
+            <h2>How Do I Install Homebrew?</h2>
```
jaimeMF: Shouldn't this part also be translated? (or make it optional)

mikemcquaid: That's the big problem we have with introducing new strings, unfortunately.

jaimeMF: It's probably not the nicest thing, but you can check if the string is available and default to the english string: https://github.com/jaimeMF/homebrew.github.io/commit/49ec3470df2631e267c9637b2709f846ffbd2964. If someone browsing in another language notices it, he can submit the translation.

jacknagel:
I see a few formatting issues, namely:

 * the horizontal rules don't all match anymore
 * the first new item seems to be inside a `<li>` tag, but the next new header is not... seems ripe for styling issues. there is also some curious indentation changes happening, in part because of this I think
 * there is a single new `<br>`, but it's the only one on the page, probably this can be accomplished by fixing the other issues

_layouts/index.html
DomT4: Yeah, new strings are difficult on a website that has already been translated so widely. I tried to keep the two strings fairly accessible in terms of word choice but I appreciate native language support is a better solution.

DomT4:
> the horizontal rules don't all match anymore and there are too many of them, it's noisy

Agreed. There's some spacing issues to be ironed out, which also should fix your third point in the temporary hack I stuck in place by inserting a hard break ` <br> ` to gain a little more space.

> the first new item seems to be inside a <li> tag, but the next new header is not... seems ripe for styling issues.

That was a slip on my part. Will be fixed :smile_cat:

DomT4:
Heh, Git closed the branch because I reset against the head and then rebased against the master to get the gem updates and then pushed to the remote forgetting to specify to push only the master. That was stupid of me. Still working on this, despite that klutz move.

DomT4:
Opened #55.

PR #55, opened 2015-02-22
rejiggle layout for additional clarity

DomT4:
[What it looks like](https://domt4.github.io/).

See #54
Closes https://github.com/Homebrew/homebrew/issues/36970

### DIFF
--- a/_layouts/index.html
+++ b/_layouts/index.html
@@ -68,6 +68,22 @@
         <ul>
           <li>
             <div class="group row">
+              <h2 id="install">{{ page.pagecontent.install.install }}</h2>
+              <br>
+              <pre style='clear:both;text-align:center;margin-bottom:0.9em'><code id='selectable' onclick="selectText(this)">ruby -e &quot;$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)&quot;</code></pre>
+              <div class="col-1">
+                <p>{{ page.pagecontent.install.paste }}</p>
+              </div>
+              <div class="col-2">
+                <p>{{ page.pagecontent.install.what }}</p>
+              </div>
+            </div>
+          </li>
+
+          <li>
+            <div class="group row">
+             <h2 id="question">{{ page.pagecontent.question }}</h2>
+             <br>
               <div class="col-1">
                 <p>{{ page.pagecontent.what }}</p>
               </div>
@@ -87,12 +103,12 @@ $ brew install wget
 {% highlight bash %}
 $ cd /usr/local
 $ find Cellar
-Cellar/wget/1.15
-Cellar/wget/1.15/bin/wget
-Cellar/wget/1.15/share/man/man1/wget.1
+Cellar/wget/1.16.1
+Cellar/wget/1.16.1/bin/wget
+Cellar/wget/1.16.1/share/man/man1/wget.1
 
 $ ls -l bin
-bin/wget -> ../Cellar/wget/1.15/bin/wget
+bin/wget -> ../Cellar/wget/1.16.1/bin/wget
 {% endhighlight %}
               </div>
             </div>
@@ -111,7 +127,7 @@ bin/wget -> ../Cellar/wget/1.15/bin/wget
               </div>
               <div class="col-2">
 {% highlight bash %}
-$ brew create http://foo.com/bar-1.0.tgz
+$ brew create https://foo.com/bar-1.0.tgz
 Created /usr/local/Library/Formula/bar.rb
 {% endhighlight %}
               </div>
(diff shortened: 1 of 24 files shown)


DomT4:
So, talked about this a little more, and pondered some of the feedback in the other thread, so changelog:

Unrelated to the problem but since I was there anyway:
* Updates all of Max's links. We'd done the English one previously, but not the others.
* Fixes a few links that weren't SSL/TLS but should be.
* Shuffles the install block towards the top where it's now displayed on the website.
* The full link at the bottom of the page for the documentation now just reads ` Homebrew @ Github `. As "trademark" names (per se) they don't need translating, so there's no extra work, and it looks a little cleaner.

Related to the problem:
* Uses inline CSS because I didn't think the addition of what is essentially 2 div blocks merited writing a whole new class in the CSS file which would add to the maintenance burden going forwards rather a lot.
* Consequent to that, we retain the normal borders of the information block but just work on top of them. You can't see them under the changes, so it's not a big thing.
* There was some understandable stress amongst having to translate the new strings, and I agree that wasn't an ideal situation so I decided to try and make the block *pop* off the page via the medium of CSS instead of actually using words.
* Install block is now at the top.
* Install block is now wrapped in Homebrew-Gold (As I'll refer to it rather than using the hexcode here) 6px block. There is a 1px padding inside the block to give it a very, very slightly 3D look. I couldn't decide between a dotted border and an outset border, but the latter provided more obvious separation so went with that.
* There's an enforced line break underneath the border to give it more separation.
* Underneath that enforced line break I've thickened the regular border and made the line black. Yet more separation effort.

Welcome thoughts from @jacknagel, @mikemcquaid and everyone else as well, obviously. If you don't love the border solution, we can try something else again :smile_cat:.

mikemcquaid:
@DomT4 I think we can introduce new text and just use the approach in #56.

bfontaine:
> Uses inline CSS because I didn't think the addition of what is essentially 2 div blocks merited writing a whole new class in the CSS file which would add to the maintenance burden going forwards rather a lot.

I respectfully disagree:

1. HTML is not the place where you’d expect to find some CSS.
2. Inline CSS takes precedence over “normal” CSS, thus preventing you from overriding it in the CSS files. It’s easy to forgot inline CSS, and when you do, you can spend hours trying to fix a problem where your changes are overridden by the inline code. This is related to my first point.

DomT4:
> I think we can introduce new text and just use the approach in #56.

WFM if you're happy with that. I'll rebase and do some work on that tonight.

> HTML is not the place where you’d expect to find some CSS.

Sure, theoretically. But in reality you tend to find CSS dotted all over HTML files. It's become and is becoming less popular, but you can certainly find enough places on the web which have chunks of inline CSS. Whilst generally it should be dumped in a specific ` .css ` file, sometimes there are merits to filing it inline, if you can keep track of where it is. You can stick a code comment above the block along the lines of ` <!-- I AM CSS. DON'T FORGET ME --> ` :smile_cat:.

mikemcquaid:
I agree on the inline CSS; better to push it to the CSS file(s) instead.

DomT4:
Alright, I don't love inline CSS enough to argue with two of you on this :wink:. Any further changes I'll outsource to the CSS files :smile_cat:.

xu-cheng:
Maybe it's a bit off topic. But can we have a more flatten design like Yosemite and iOS 7 did.

mikemcquaid:
@xu-cheng I don't really want to mess with the design too much for now.

xu-cheng:
I'm not web developer. And that's just a general suggestion. So feel free to ignore it completely. :wink:

mikemcquaid:
Needs rebased.

_layouts/index.html
```diff
@@ -147,22 +165,9 @@
...
-              </div>
-            </div>
-          </li>
-
-          <li>
-            <div class="group row">
               <h2 id="further-doc">{{ page.pagecontent.doc.further }}</h2>
               <div class="button">
-                <p><a href="https://github.com/Homebrew/homebrew/tree/master/share/doc/homebrew#readme">https://github.com/Homebrew/homebrew/tree/master/share/doc/homebrew#readme</a></p>
+                <p><a href="https://github.com/Homebrew/homebrew/tree/master/share/doc/homebrew#readme">Homebrew @ Github</a></p>
```
mikemcquaid: After rebasing this link has been shortened so let's leave it as-is.

index_ar.html
```diff
@@ -5,6 +5,10 @@
 direction: rtl
 
 pagecontent:
+  install:
+    install: Homebrew ثبّت
```
mikemcquaid: How did you get all these?

mikemcquaid: Ah, duh, they just got moved up. Why did they get moved?

DomT4:
> Needs rebased.

Already done locally, have just been working on it a bit more, so hadn't pushed yet :).

_layouts/index.html
DomT4: Nice, thanks.

index_ar.html
DomT4: > How did you get all these?

I am a master of all languages :wink:. *(I wish)*.

>  Why did they get moved?

It seemed more canonical given the new location of the install block on the page. I was possibly just being anal though.

mikemcquaid: Yeh, let's just leave them as-is.

DomT4: Cool. Done locally. Will give it a push up to the remote here within the next couple of hours.

DomT4: > Will give it a push up to the remote here within the next couple of hours.

Or I would have done if I hadn't suddenly fallen asleep last night :sweat_smile:. Just finishing up now.

DomT4:
Pushed and ready for critique.

mikemcquaid:
Thanks @DomT4!

DomT4:
Thanks Mike, and everyone else for the feedback on this. Appreciate it all :+1:

PR #60, opened 2015-02-26
use https://git.io/brew-docs/ to redirect document.

xu-cheng:
* Create the webpage http://brew.sh/doc/ which will redirect to https://github.com/Homebrew/homebrew/tree/master/share/doc/homebrew#readme.
* Use http://brew.sh/doc/ to short document url.

### DIFF
--- a/_layouts/index.html
+++ b/_layouts/index.html
@@ -162,7 +162,7 @@ end
             <div class="group row">
               <h2 id="further-doc">{{ page.pagecontent.doc.further }}</h2>
               <div class="button">
-                <p><a href="https://github.com/Homebrew/homebrew/tree/master/share/doc/homebrew#readme">https://github.com/Homebrew/homebrew/tree/master/share/doc/homebrew#readme</a></p>
+                <p><a href="https://git.io/brew-docs">https://git.io/brew-docs</a></p>
               </div>
             </div>
           </li>


mikemcquaid:
Instead of the redirect lets use http://git.io/brew-docs instead.

xu-cheng:
I don't think http://git.io/brew-docs is a good idea. As it's hard linked to https://github.com/Homebrew/homebrew/tree/master/share/doc/homebrew#readme. If by any chance in the future, we want to move the documents somewhere else(I know it's slightly in foreseeable future), we would have to change the link again. On the contrary, using http://brew.sh/doc/ a webpage we can control, we can redirect it to anywhere if we want.

mikemcquaid:
I don't really like making technical decisions based on speculative unknown future decisions. Again, I don't think that's solving a problem users are experiencing. We can add a new git.io link if we want in future or I can change it. The Jekyll redirects are a bad hack and nasty user experience.

xu-cheng:
Isn't http://git.io/brew-docs also a redirect?

>  I can change it

Really? sweet power. :heart_eyes: 

OK, I will change into http://git.io/brew-docs.

mikemcquaid:
It's a 301 redirect rather than a HTML one. With the latter some people in some browsers will see a flickering as the other page briefly loads.

xu-cheng:
Updated.

xu-cheng:
@mikemcquaid I wonder can we use the same tech to short the install url.

mikemcquaid:
@xu-cheng Probably not; you need to pass an additional argument to e.g. `curl` to get it to follow redirects.

xu-cheng:
@mikemcquaid I know that part. And we have the argument (`-L`) in curl command. 

The reason I don't think it's a good idea using `git.io` is I thought we cannot control it (for example relocate it) after creating the link.

PR #61, opened 2015-03-10
Small text translation missing in pt-br version of home page

joelwallis:
Translating the subtitle *What Does Homebrew Do?* to Brazilian Portuguese.

Also, I could see that it's not translated in all of other languages. What's the point in keeping this subtitle in English?

### DIFF
--- a/index_pt-br.html
+++ b/index_pt-br.html
@@ -4,7 +4,7 @@ lang: pt-BR
 subtitle: O gerenciador de pacotes que faltava para o OS X
 
 pagecontent:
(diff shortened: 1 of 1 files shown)


mikemcquaid:
> Also, I could see that it's not translated in all of other languages. What's the point in keeping this subtitle in English?

We introduced it as a new string so have to wait for it to be translated by kind people such as yourself, unfortunately.

joelwallis:
Oh, really? So let me ask for some help from multilingual friends and make a thousand more PRs, hehe.

PR #64, opened 2015-03-22
Slightly improved German index file

denpatin:
* Corrected one misspelling


mikemcquaid:
Thanks @denpatin!

PR #65, opened 2015-03-22
Slightly improved Italian index file

denpatin:
* Corrected the adjective form (I presume that it is not a manager that is missing but packages)


mikemcquaid:
Thanks @denpatin!

PR #67, opened 2015-03-22
Slightly improved Russian index file

denpatin:
* Corrected the `subtitle` section


mikemcquaid:
Thanks @denpatin!

PR #71, opened 2015-03-23
Slightly improved Belarusian index file

denpatin:
* Translated the `editor` section, which I forgot to translate
* Slightly corrected the misspelling


mikemcquaid:
Thanks @denpatin!

PR #74, opened 2015-04-07
Smoothen Font

snarkDoe:
This style looks better and easier to read


mikemcquaid:
I'd rather leave it as-is, thanks. Sorry!

PR #77, opened 2015-05-06
Add missing French translation in index_fr.

deansc:
Just updated a missing translation in index_fr.html file.


xu-cheng:
Thanks @deansc  :tada:

Issue #41, opened 2014-10-03
The font looks different for latin and not-latin characters


romcheg:
For non-latic characters that are used in titles this font is not as bold as for latin ones. Since some titles contain both latin and non-latic characters, this looks odd.
![screen shot 2014-10-03 at 16 40 06](https://cloud.githubusercontent.com/assets/214072/4507327/a2f3da52-4b0b-11e4-8097-2e3226346ead.png)
![screen shot 2014-10-03 at 16 40 26](https://cloud.githubusercontent.com/assets/214072/4507328/a2f53a82-4b0b-11e4-8ae3-924c3951eecd.png)

mikemcquaid:
I think this is a WONTFIX I'm afraid due to the desire to maintain our branding with the current font. Sorry!

PR #81, opened 2015-08-31
Added Swedish translation

WizuO:
Added Swedish translation strings to brew.sh website.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -35,6 +35,8 @@ langs:
     lang_string: Português Brasileiro
   - langcode: ru
     lang_string: Русский
+  - langcode: se
+    lang_string: Svenska
   - langcode: th
     lang_string: ไทย
   - langcode: tr
--- a/index_se.html
+++ b/index_se.html
@@ -1,0 +1,25 @@
+---
+layout: index
+lang: se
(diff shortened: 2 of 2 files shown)


mikemcquaid:
Thanks!

WizuO:
No problem! Just heads up, the 'ä' character in one of the titles isn't bolded right now. Does it need any extra html tags? 
Link: http://brew.sh/index_se.html

dunn:
My guess is that's a limitation of the webfont we're using.

WizuO:
If needed I can just revert that specific title back to English, as I noticed the Norwegian translation has done. There is really no other way of writing it without using non-English characters.

dunn:
I'd be open to switching to a new font, if we can find a similar free one with better character support.

bjorno:
Yeah, that could be necessary. I just realized this issue has already been brought to attention, with no resolution: https://github.com/Homebrew/homebrew.github.io/issues/41

dunn:
I'll open a PR soon to switch to this font https://www.google.com/fonts/specimen/Roboto+Slab unless we come across a better alternative.

bjorno:
Looks like a good replacement. Thanks you!

PR #84, opened 2015-09-15
slight fixes on Japanese translation

polamjag:
- modify some particles (助動詞)
- append space between Japanese and English word


xu-cheng:
Thanks for the pull request! 🎉 Homebrew depends on contributions from community members like you and we're grateful for your support.

PR #100, opened 2016-03-14
Added Danish translations

praffn:
The webfont (Chunkfive) seems to not support the Ø and Å glyphs, which are extremely prevalent in Scandinavian languages, which is why the title "What Does Homebrew Do?" is still in English, like the Swedish and Norwegian translations – it's simply impossible to translate that sentence in a meaningful way without those glyphs 🐭


mikemcquaid:

PR #115, opened 2016-05-08
Typos

retokromer:
In French as used in France before the punctuation ! ? : and ; there is a non-breaking space. I unified this.


retokromer:
Oops, wrong button.

mikemcquaid:
Thanks again @retokromer!

PR #117, opened 2016-06-22
OS X is now macOS

Chris54721:
```sh
sed -i -- 's/Mac OS X/macOS/g' *.html
sed -i -- 's/OS X/macOS/g' *.html
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,7 +1,7 @@
 ---
 layout: index
 lang: en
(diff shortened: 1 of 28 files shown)


dunn:
El Capitan is still called OS X (http://www.apple.com/osx/) so we should probably wait on this until Sierra is actually released.

PR #118, opened 2016-08-01
Translated remaining title with non-standard characters

bjorno:
The remaining title needed non-standard characters to be translated. With the new font support it can finally be implemented.


MikeMcQuaid:

PR #119, opened 2016-08-04
Added partial finnish translation

theel0ja:
Some of translations are partially compelete, so improvement is needed.


MikeMcQuaid:

PR #120, opened 2016-08-05
added finnish to config

theel0ja:


theel0ja:
releated to #119

MikeMcQuaid:
Thanks again @theel0ja!

PR #121, opened 2016-08-20
OS X -> macOS

ZakariaRidouh:
This also fixes some inconsistencies throughout such as 'Mac OS X', 'OS X', 'OSX'.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,7 +1,7 @@
 ---
 layout: index
 lang: en
-subtitle: The missing package manager for OS X
+subtitle: The missing package manager for macOS
 
 pagecontent:
   question: What Does Homebrew Do?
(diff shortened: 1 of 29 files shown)


MikeMcQuaid:
Apple still calls it OS X so we will too for now: http://www.apple.com/uk/osx/

ZakariaRidouh:
Hey!

According to Wikipedia:

> The latest software version is OS X El Capitan, which was publicly released in September 2015. In June 2016, Apple announced the rebranding of the software to macOS starting with the upcoming release, macOS Sierra ...


You've gone with OS X as that was the name last year. Since Apple has now officially rebranded OS X to macOS, I think you should update the usage everywhere throughout the documentation.

Alternatively, it is possible to just wait until Apple updates their website to macOS and then you can reopen this.

Cheers 😄 !

bfontaine:
It'll only take effect when Sierra is released (emphasis mine):

> Apple announced the rebranding of the software to macOS _starting with the upcoming release_

PR #122, opened 2016-08-24
Update github corner

vinkla:
Switched to the new better looking github corners http://tholman.com/github-corners/ I've added the red corner.


MikeMcQuaid:
I prefer the old one, I think it's more obvious what it's doing if you don't recognise the :octocat:.

vinkla:
Yeah, I agree the old one is more obvious. Though, IMO the new one looks so good.

MikeMcQuaid:
I do quite like the new one but probably more for projects where the chances of people not knowing the :octocat: are 0️⃣. Let's leave this open for a bit and see what others think.

vinkla:
Okay, sounds fare.

UniqMartin:
I'm afraid I have to agree with Mike that I also prefer the old one. The new ones are minimalist and cute, but a lot less obvious. However, if we decided to pick one of the new ones, I think we should adjust the colors to match our color scheme.

vinkla:
@UniqMartin which color do you prefer?

UniqMartin:
I think the Octocat using the background color and the corner itself using the text color of the headings (like `h1` and `h2`).

vinkla:
@UniqMartin sounds good. It is fairly easy to add custom colors. We can just update the hex value.

MikeMcQuaid:
Passing, sorry!

vinkla:
No worries!

PR #130, opened 2016-09-21
Fix html validation errors

teddyzetterlund:
I [ran the site through W3C validator](https://validator.w3.org/nu/?showimagereport=yes&doc=http%3A%2F%2Fbrew.sh%2F) and noticed two simple errors:

1. Duplicate `<title>` elements due to the SEO plugin, which is unfortunate when there's a beautiful display of multilingualism going on.

2. Non-digits inside the `width` and `height` attributes of an `img` element for the logo.

Cheers 🎉

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -10,7 +10,7 @@
     {% else %}
     <title>Homebrew — {{ page.subtitle }}</title>
     {% endif %}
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:

PR #134, opened 2016-09-25
Update code highlighting color scheme and ditch Pygments

remiprev:
So I promised to revisit the syntax highlighting color scheme because I think it clashes with the rest of the site. I also removed Pygments because I think we don’t need it since we control the highlighted code content and it doesn’t change very often 😄 

#### After

Let me know what you think!


MikeMcQuaid:
I'd rather we still relied on Rouge/Pygments but I'm definitely open to changing the colour scheme.

remiprev:
No problem, I’ll close this PR and open up a much simple one!

PR #135, opened 2016-09-25
Update Pygments color scheme

remiprev:
_These are the same changes as #134 but without Pygments/Rouge removed_ 😄 

***

So I promised to revisit the syntax highlighting color scheme because I think it clashes with the rest of the site.

#### After

Let me know what you think!


MikeMcQuaid:
Nice work 👍

PR #138, opened 2016-09-25
Add hreflang attribute links to head

Eitot:
This adds `<link>`s with attribute `hreflang` to `<head>`. Search engines such as Google support this attribute and will use it for indexing and localising search results. Perhaps this improves the discovery of the language-specific page in the search results.

Reference: https://support.google.com/webmasters/answer/189077

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -39,6 +39,14 @@
       ga('send', 'pageview');
 
     </script>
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Makes sense. Thanks again @Eitot!

PR #143, opened 2016-10-14
Remove outer margin from <figure>

Eitot:
This removes the right margin from the `<figure>` elements (used for the command examples) to make it even with the top.

screen shot 2016-10-15 at 00 43 14

### DIFF
--- a/css/screen.css
+++ b/css/screen.css
@@ -184,6 +184,14 @@ pre code {
   padding: 0 0 1em;
 }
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Can I see the before/after screenshot? Thanks!

Eitot:
screen shot 2016-10-15 at 13 39 00

MikeMcQuaid:
Gotcha, much nicer, thanks!

PR #152, opened 2017-02-27
fix https mixed content issue

JCount:
the current url for the fork-me img resource is transparently
redirecting to http, so fix the url to correctly use aral.github.io


MikeMcQuaid:
Nice spot, thanks @JCount!

Issue #154, opened 2017-03-14
Links on 'blog page, to individual posts, are not visually apparent


matatk:
On https://brew.sh/blog/ there are a series of post titles and abstracts, where the post titles link to the full text of each 'blog post.

However, there are no visual affordances that the posts' titles are actually links to the full posts, or that there is any more of the post to read than the initial paragraph.  An example of this is the post announcing v1.0.0 - it contains quite a bit of detail about the release, but it's not clear that all that information is there, because the heading of the post is not visually styled like a link, and there's no ~"read the full v1.0.0 article" link below the first paragraph.

Just having a "read more" link below each article's abstract would not be good for accessibility, because people using assistive technologies to browse the web often navigate through the links in a page (e.g. using <key>Tab</key>, or via a list of all links on the page) and in that case they'd just be presented with a series of identical "read more" links, rather than "read more about the v1.0.0 release" and so-on.  It is possible to include extra text in links to give such users the context of the link, without having it appear visually, as per [Example 1 in the Web Content Accessibility Guidelines (WCAG) Technique for hiding some link text with CSS](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/C7).

If you would like to address this, then a good simple approach would be to include such "read more about X" links (with sufficient context for users of assistive technologies, as above) rather than having the headers for each post as links, as doing so overloads their purpose (generally, heading elements are for navigation within the current page; links are for navigation to other pages).  However, if the heading links are to be kept, styling them in the same way as the other links on the site would be appropriate, so at least users know they're links.

A further, related issue [if you like I can file this separately] is that the contrast between link text and the page background is only 4.3:1, whereas 4.5:1 is recommended by WCAG as a minimum, and the contrast between body (non-link) text and link text is only 2.3:1, whereas at least 3:1 is a good minimum (and 4.5:1 is preferable if possible). One solution to this would be to lighten the link text a bit to get it to 4.5:1 and also have links visually underlined, so that even people who can't distinguish them from the body text by colour will still be able to tell they're links, due to the underline.

MikeMcQuaid:
> rather than having the headers for each post as links, as doing so overloads their purpose (generally, heading elements are for navigation within the current page; links are for navigation to other pages)

I don't agree with this, I'm afraid. I don't think I've seen many mainstream blogs where the headers for each post _aren't_ links. It may well be they are doing it wrong too, I admit.

[...]

PR #160, opened 2017-04-27
Contrast for links and links-vs-body text

matatk:
This follows on from the discussion in https://github.com/Homebrew/homebrew.github.io/issues/154

I have been using http://jxnblk.com/colorable/demos/matrix/ to tweak the colours of:

 * the page background;
 * body text and
 * link text

so that they are, hopefully, easier to read - particularly so that links are more distinguishable from the body text.  Here is my first effort at this.

Here are the original colours and contrast ratios:
homebrew-original
Of note:
 * Link text to page background is 4.33:1
 * Link text to body text is only 2.27:1

Here are the ones I currently have:
homebrew-proposed
Of note:
 * Link text to page background is now only 4:1, which is not as good, though it still passes WCAG 2 Level AA under the "large text" exception, which states that if the text is large (which this qualifies as, due to being bold, on top of the larger base font size the site as a whole uses), then it's acceptable.
 * Link text to body text is now 3:1, which means more people are going to be able to differentiate between them.

I've been fiddling with the sliders for some time, but couldn't come up with a palette that felt the same as the original one that doesn't also have a slightly worse link text to background ratio.  In fact I am now somewhat inspired to write some code to exhaustively find the best such ratios between three colours, with artistic constraints, but that's probably not happening anytime soon...

Preserving the original character of the site was a priority and I hope you feel I've succeeded in doing this.  I'm still a little reticent due to the reduction in contrast between link text and page background (even though it's still considered acceptable under Level AA, it is worse than what was there originally) so might have another go with these sliders over the next few days again, to see if I can do any better.

A "wildcard" alternative approach, which could support even higher contrast throughout, would be to use a very different hue for the link colour, such as green (I think it looks quite cool, but imagine many may not!)

This PR also makes the following changes:

 * The new body text colour is used for headings too (so they match)
 * The new link text colour is used as the selection background colour (to simplify the palette)

### DIFF
--- a/css/screen.css
+++ b/css/screen.css
@@ -21,8 +21,8 @@ html, body { height: 100%; }
 body {
   font-size: 150%;
   line-height: 1.4;
-  color: #F9D094;
-  background: #2E2A24;
+  color: #ffdead;
+  background: #25241d;
   position: relative;
   behavior: url("/js/ie6/csshover.htc");
   padding: 0 30px;
@@ -30,10 +30,10 @@ body {
 
 p,ul,ol,dl,table,pre { margin-bottom: 1em; }
 ul { margin-left: 20px; }
-a { text-decoration: none; cursor: pointer; color: #ba832c; font-weight: bold; }
+a { text-decoration: none; cursor: pointer; color: #ba700b; font-weight: bold; }
 a:focus { outline: 1px dotted; }
 a:visited {  }
-a:hover, a:focus { color: #d3a459; text-decoration: none; }
+a:hover, a:focus { color: #ffdead; text-decoration: none; }
 a *, button * { cursor: pointer; }
 hr { display: none; }
 small { font-size: 90%; }
--- a/css/screen.css
+++ b/css/screen.css
@@ -46,8 +46,8 @@ button::-moz-focus-inner { border: 0; padding: 1px; }
 span.amp { font-weight: normal; font-style: italic; font-size: 1.2em; line-height: 0.8; }
 h1,h2,h3,h4,h5,h6 { line-height: 1.1; }
 
-::selection { background: #745626; }
-::-moz-selection { background: #745626; }
+::selection { background: #ba700b; }
+::-moz-selection { background: #ba700b; }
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Will give some more lengthy thoughts on this soon but CC @remiprev for thoughts on this; he came up with the original palette and has more aesthetic sense in his toe than I have in the rest of my body combined 😆

remiprev:
@matatk Could you post a screenshot of the whole page _before_ and _after_ your code changes? I don’t mind improving the colors, I’d just like to compare the end result.

MikeMcQuaid:
@matatk I'm afraid I merged https://github.com/Homebrew/homebrew.github.io/pull/166 which will cause a bunch of nasty merge conflicts such that it's maybe worth trying to remake this PR from scratch again. As a positive you should now only need to change colours in a single location.

matatk:
@MikeMcQuaid I've been a bit laggy on this; no worries. Will have a look and possibly re-do. Nice to hear regarding the DRY on the CSS :-).

@remiprev sure; will send when I've updated/redone this to fit the new CSS.

I'm still somewhat torn by the fact that, in order to bring up the contrast between link and body text, I had to reduce the contrast of the body text (both are within acceptable levels for large fonts at WCAG level AA, though it feels like it shouldn't be a zero-sum game—though the only way to increase *both* would be to use a different accent/link colour for the site, which would change the appearance a lot). Anyway, more on that in a re-worked PR :-).

MikeMcQuaid:
@matatk I can't remember if I mentioned this previous but another option for finding colours (and prompted the new colour scheme) are these new icons: 
https://avatars2.githubusercontent.com/u/1503512?v=3&s=400
https://avatars1.githubusercontent.com/u/1589480?v=3&s=400

matatk:
@MikeMcQuaid: thanks for the pointers to the artwork. Sorry I have not yet done the other PR; it's a simple patch, but I have been super-busy; will be on it soon, though.

matatk:
Have created #167 now, so will close this.

PR #161, opened 2017-04-28
analytics/index: fix typo

JCount:
fix analytics/index.html typo:
`(i.e. no pulled in as a dependency)` to `(i.e. not pulled in as a dependency)`


MikeMcQuaid:
Thanks @JCount!

PR #164, opened 2017-05-02
Link analytics formulas to braumeister

joshka:
I was looking at the shiny new analytics pages and wondering what a particular package did. To my dismay I had to search manually rather than clicking. This fixes that oversight.

### DIFF
--- a/_layouts/analytics.html
+++ b/_layouts/analytics.html
@@ -14,7 +14,11 @@ layout: base
 {% for item in json.items %}
   <tr>
     <td class="number-data">#{{ item.number }}</td>
-    <td><code>{{ item.formula }}</code></td>
+    <td>
+      <a href="http://braumeister.org/formula/{{ item.formula }}">
+        <code>{{ item.formula }}</code>
+      </a>
+    </td>
     <td class="number-data">{{ item.count }}</td>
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Seems sensible, thanks!

MikeMcQuaid:
Although: a bunch of these don't exist: https://travis-ci.org/Homebrew/homebrew.github.io/builds/227853868?utm_source=github_status&utm_medium=notification so we probably need to think about how to handle that more gracefully than a 404.

joshka:
Good thing there's external link tests :)

Probably the long term correct way to handle this is include something like braumeister directly on brew.sh down the track.

A quicker workaround might be to omit the link if the package isn't on braumeister, by adding the url check to the analytics data generation script (or a later script).

MikeMcQuaid:
> Probably the long term correct way to handle this is include something like braumeister directly on brew.sh down the track.

That's never going to happen because `brew.sh` is a static site and braumeister dynamic. Braumeister may get on our subdomains eventually though.



> A quicker workaround might be to omit the link if the package isn't on braumeister, by adding the url check to the analytics data generation script (or a later script).

Yeh, was thinking the same. Doing this generation time probably makes most sense.

PR #167, opened 2017-06-23
Contrast

matatk:
Right, got ‘round to this at last.  I decided to take the [new icon](https://avatars1.githubusercontent.com/u/1589480?v=3&s=400) that @MikeMcQuaid shared as the starting point, so that the colours come from something more authoritative.

I applied the same approach to both of the colours in the icon (the beer and the beer highlight), to see which worked out best:

 * Start with the base colour.  From this, produce:
   - A slightly darker version, for link text.
   - A lighter version, for body text.
   - A much darker version, for the page background (if you want to change that).
 * Adjust until the contrast between…
   - …link and background is at least 3:1.
   - …body text and background is at least 4:1 (it needs to be as high as possible because, whilst I’m counting this as “large text” because it’s bold, it’s not all that big relative to the body text).

I could only make this work with the colours that arose from the “beer” colour in the icon (the ones that came from “beer highlight” didn’t seem as warm, anyway).

I’ve attached an image that shows all the ratios (created with [colorable](http://jxnblk.com/colorable/demos/matrix/)), and another two with the Homebrew home page with the previous and suggested colours.

There are two issues:

 1. The background is now quite a bit darker, which is noticeable; I could repeat this process and try to make the background colour match a bit better, perhaps by adjusting the saturation as well as the lightness, for the background colour.

 2. The contrast between link text and background is low; again I could have another go at this process, and try to sort that too.

As this is probably just the start, I thought I’d send the results so far.

homebrew-contrast-process
homebrew-original
homebrew-proposed

Fixes https://github.com/Homebrew/homebrew.github.io/issues/154


MikeMcQuaid:
This is looking great. It's almost perfect from my perspective bar two things:
- I'd like the background to be a little more "brown" and lighter. If that means the links colour changes: that's cool.
- Tweak the separator colours (it's basically two border colours combined) so they are a bit more visually prominent. Arguably they could be a bit more prominent in the original but they get totally lost now unless you're actively looking for them.

matatk:
I've just added another commit - the message explains it, but essentially I started with the original background and just darkened it a bit, and tweaked the two new highlight colours (though they're both still from the main colour in the new icon).  I think this is a bit better:

homebrew-site-patch-r2

I think this might be getting closer, but:

1. Should the separator colours still be tweaked? I think maybe a bit, yes?
2. Should I rename any of the colour variable names in the CSS, as they're not quite what they were (and, if so, do you prefer colour names or their primary function e.g. `link_text`)?

MikeMcQuaid:
The new background feels even less brown (which I think is the main essence of Homebrew's colour scheme). I know it's really annoying but I feel like it's now preferably to seek to keep the background the same and change the e.g. text and/or link colour to add sufficient contrast. I'd also consider underlining all links if it means maintaining a similar colour scheme at this point.

matatk:
That's fine... let me just check to be sure: you want to stick with the original/current background colour (`#2e2a24`) and then adjust the links, and text colours from there? I think the ones that came from the beer colour in the new icon were pretty nice (warm), and they should go fine on the new background (they're already WCAG-compliant, as per the image [in the first post above](https://github.com/Homebrew/homebrew.github.io/pull/167#issue-238128610), but with a little tweaking I think they could be a lot better; will have a go...

MikeMcQuaid:
Yeh, that sounds good. At least it would be good to have a look at that approach.

> #2e2a24

Other previous options were `#27221a` and https://github.com/Homebrew/homebrew.github.io/blob/0dbf0bdf6728e82ac04d7eb8ca471021ad373f52/img/bg-body.gif in case either of those work for you.

Sorry for all the back and forth; because I'm not a designer I can only really 👍 / 👎 things when I see them.

matatk:
No worries; it needs to be right, and seeing it is absolutely necessary.

I've just tried the new text/link colours this PR proposes, on the original. Here's the [colorable](http://jxnblk.com/colorable/demos/matrix/) screen grab:

homebrew alternative colours

Here's what the current background (`#2e2a24`) with latest proposed colours (based on the beer in the new icon):

current-bkg

Here's the same with the other background colour you mentioned (which is the same as in the .gif) (`#27221a`):

new-bkg

MikeMcQuaid:
Do either/both of them meet all the necessary accessibility/contrast requirements? Thanks again!

matatk:
I've checked the desktop viewport and it does, yes;  as it qualifies as 'large text' it's subject to a 3:1 minimum requirement (though, really, the bigger the better).  I haven't looked at the other media queries in detail (though from glancing at the changes in a browser, it seems the fonts stay the same)

MikeMcQuaid:
> Here's what the current background (#2e2a24) with latest proposed colours (based on the beer in the new icon):

This looks great to me 👍

MikeMcQuaid:
> Here's what the current background (#2e2a24) with latest proposed colours (based on the beer in the new icon):

To be more explicit: assuming this passes your other tests, let's move forward with the PR given those colours you've suggested and I'll handle the e.g. variable names.

matatk:
I’ve had a more detailed look at this; here is my attempt at summarising it concisely :-).

## Background
* There are two levels of requirement for contrast ratio in WCAG Level AA:
   - 4.5:1 is required for normal text on its background.
   - 3:1 is a lower requirement for text on a background, when text is large.
* Large text is either:
   - 18-point/24px
   - 14-point/18.5px and bold

(Source: [W3C](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html)—refer in particular to the first note in green.)

## Summary
Unfortunately I got mixed up with px vs pt in my previous assessment, sorry, so currently the desktop site does not meet WCAG AA requirements for contrast, because:
* whilst body text to background passes easily (either colour scheme (original or slightly darker background is well over 10:1),
* link text to background doesn’t pass because it’s >=3:1 but is only 15px bold and not 18.5px bold, so it classifies as normal text and should be >=4.5:1.

I checked the mobile portrait layout on my ‘phone and it came out as 15px too, so would have the same result.

I checked the mobile landscape layout on my ‘phone and it passes, for both body and link text, because all the text there is 24px (or bigger).

There is no solid requirement for the ratio between link and body text, but I recommend 3:1 as a minimum, as that gives the best chance of most users being able to discern the difference.

## Possibilities
Once again, sorry I missed this earlier.  The challenge is that the link text needs to contrast more with the background, but it is difficult to do that _and_ have both the body and link text looks similar to how they do now, because the main (text) colour and highlight (link) colour are similar.  I can think of three options:

* Have another go and make the background darker and main text paler (but try to avoid making it too pale).
* Spread the site's colour palette out a bit so that instead of all three main colours coming from the same part of the spectrum, one (e.g. the accent) is a bit more spread out. Of course this would change the look of the site.  (I remember seeing a predominantly brown site once, I think it was for a JavaScript library, and the accent was green, which may not be everyone's taste, but it seemed pretty cool to me; can't find it now, though, alas.)
* Make the text larger generally (it looks like it may be sufficient to remove the 62.5% on the html, and maybe then the 150% on the body could be toned down, even).

MikeMcQuaid:
> make the background darker
> Spread the site's colour palette out a bit

At this point I'm 👎 on changing to a different colour scheme. The options as I see it at this point:

- adjust the colours so they match the current theme and are better even if not compliant
- add underlines to non-header links which will make more of them compliant

matatk:
That's fine (I actually do quite like the look of the current site, and would like to preserve its character; it just takes a little more tweaking due as an artefact of the colours all being in the same part of the spectrum).

There is a third option to add to the ones you mention: make the text bigger/bolder. Is that something you would be happy doing?  I appreciate that may change things a bit too, but just wanted to check that it was discounted before I make any further colour tweaks.

MikeMcQuaid:
> make the text bigger/bolder

Which text? Everywhere on the site? The body text? Just link text? I don't think I'm in favour of any of them but I could be convinced.

matatk:
> Which text? Everywhere on the site? The body text? Just link text? I don't think I'm in favour of any of them but I could be convinced.

[ I'm giving the media queries names here, just because it's easier to talk about them... ]

The size increase would affect link text on the desktop and portrait phone layouts.  The landscape phone layout already has 'large' text throughout, so is fine.  [ This is a quick recap of [the summary part of the above post](https://github.com/Homebrew/homebrew.github.io/pull/167#issuecomment-313924527). ]

The current link text on the two affected layouts is 15px bold, whereas to be classed as 'large' text it would need to be ~18.5px bold.

Of course, just making the link text bigger would look really strange, so in reality the answer to your question is: both the body and link text, on the desktop and phone (portrait) layouts.

Just thought I would mention it as an alternative to pushing the colours further apart, as it would preserve the current colours, but just make the text bigger.  In fact, here is a mock-up of the difference, so you can actually see it—the following is the result of some very basic hacking around with the CSS, but should give an idea; only the body and link text changes size (and the code, to match, which should also help with the contrast ratios in there)...

0-normal

1-larger

MikeMcQuaid:
> The current link text on the two affected layouts is 15px bold, whereas to be classed as 'large' text it would need to be ~18.5px bold.
> 
> Of course, just making the link text bigger would look really strange, so in reality the answer to your question is: both the body and link text, on the desktop and phone (portrait) layouts.

Ok, cool. That may be workable but there will still be some text we'll need to keep smaller (e.g. the install line should be fully viewable).

Do you have tools you're using for all these checks e.g. https://khan.github.io/tota11y/? At this point it may be quicker for me to try and give it a go myself.

matatk:
Ah, I've met tota11y before; just had a more detailed look at it and it does indeed seem very comprehensive. I am not sure if there are some circumstances in which it may be unable to detect contrast issues, though it looks like it's using an elegant and robust method.

I had been using [colorable](http://jxnblk.com/colorable/demos/matrix/) for creating new palettes; I had just noticed the link/body and link/background issues "by eye".  Happy to have a look at the wider site and/or make those font size (or colour) changes if you like—whichever you prefer.

PR #175, opened 2017-08-03
Drop a reference to macOS 10.5, which isn't supported by homebrew

alex:


MikeMcQuaid:
Thanks @alex!

alex:
@MikeMcQuaid no problem! FYI, it looks like a few of the translated index.html's also have this reference.

MikeMcQuaid:
@alex Would you mind removing it from them too? Feel free to say "no" and I'll take a look. Thanks again!

alex:
@MikeMcQuaid #176, cheers

PR #176, opened 2017-08-05
Removed references to macOS 10.5

alex:
Refs #175 

I speak hardly any of these languages, so this was best-effort.


MikeMcQuaid:
Great work, thanks again @alex!

PR #181, opened 2017-09-17
Minor fixes (proposal)

kant:


index_es.html
```diff
@@ -25,6 +25,6 @@
     analytics: Analytics Data
```
MikeMcQuaid: Any chance you could translate these lines above? Thanks!

MikeMcQuaid:
Thanks @kant!

PR #182, opened 2017-10-25
css: fix poor visibility in syntax highlighting

jonchang:
Some syntax highlighting on docs.brew.sh are illegible due to low contrast.

Before:

image

After:

image

Blocks https://github.com/Homebrew/brew/pull/3356


MikeMcQuaid:
Thanks again @jonchang!

PR #183, opened 2017-11-04
Fix translation so it actually makes sense

mailopl:
The translation wasn't making sense in polish language. This fixes it.


MikeMcQuaid:
Thanks for your first contribution to Homebrew, @mailopl! Without people like you submitting PRs we couldn't run this project. You rock!

PR #194, opened 2017-12-22
Update _config.yml

B1NAR10:
add new language (Galician)


B1NAR10:
Sorry it's my first contribution to a project and I still don't  dominate git

MikeMcQuaid:
@B1NAR10 That's OK but cannot accept this change as-is I'm afraid.

PR #202, opened 2018-01-26
Russian subtitle translation fix

ilyamodder:
Original russian subtitle translates as "manager of missing packages", not "missing package manager"


reitermarkus:
Thanks!

PR #209, opened 2018-02-27
Add Galician (Galego) to homebrew homepage locales

anuf:
I have added translations for Galician (gl for ISO 639-1) made manually (not on transifex). More info about the language [here](https://en.wikipedia.org/wiki/Galician_language)


MikeMcQuaid:
Thanks so much for your contribution! Without people like you submitting PRs we couldn't run this project. You rock, @anuf

PR #210, opened 2018-02-28
Fix locale to ISO 639-1 in index_gl.html

anuf:
index_gl.html was submitted with a typo. Now its fixed and working.


MikeMcQuaid:
Thanks so much for your contribution! Without people like you submitting PRs we couldn't run this project. You rock, @anuf

PR #228, opened 2018-06-18
Add `brew cask` content to homepage.

MikeMcQuaid:
Add the main contents of https://caskroom.github.io to https://brew.sh.

This should allow the deprecation and redirection of https://caskroom.github.io.

screen shot 2018-06-18 at 20 16 23

CC @Homebrew/cask maintainers for thoughts on this and what you think are the hard blockers on migrating https://caskroom.github.io to here.

### DIFF
--- a/_layouts/index.html
+++ b/_layouts/index.html
@@ -9,6 +9,7 @@ layout: base
         <h2 id="install">{{ t.pagecontent.install.install }}</h2>
         <br>
         <pre style='clear:both;text-align:center;margin-bottom:0.9em'><code id='selectable' onclick="selectText(this)">/usr/bin/ruby -e &quot;$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)&quot;</code></pre>
+
         <div class="col-1">
           <p>{{ t.pagecontent.install.paste }}</p>
         </div>
@@ -32,6 +33,7 @@ $ brew install wget
         </div>
       </div>
     </li>
+
     <li>
       <div class="group row">
         <div class="col-1">
@@ -51,6 +53,7 @@ bin/wget -> ../Cellar/wget/1.16.1/bin/wget
         </div>
       </div>
     </li>
+
     <li>
       <div class="group row">
         <div class="col-1">
@@ -58,6 +61,7 @@ bin/wget -> ../Cellar/wget/1.16.1/bin/wget
         </div>
       </div>
     </li>
+
     <li>
       <div class="group row">
         <div class="col-1">
@@ -71,6 +75,7 @@ Created /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/bar.rb
         </div>
       </div>
     </li>
+
     <li>
       <div class="group row">
         <div class="col-1">
@@ -83,6 +88,7 @@ $ brew edit wget # {{ t.pagecontent.editor }}
         </div>
       </div>
     </li>
+
     <li>
       <div class="group row">
         <div class="col-1">
@@ -104,10 +110,38 @@ end
         </div>
       </div>
     </li>
+
     <li>
       <div class="group row">
         <div class="col-1">
           <p>{{ t.pagecontent.complement }}</p>
+        </div>
+      </div>
+    </li>
+
+    <li>
+      <div class="group row">
+        <div class="col-1">
+          <p>{{ t.pagecontent.caskinstall }}</p>
+        </div>
+        <div class="col-2">
+{% highlight bash %}
+$ brew cask install firefox
+{% endhighlight %}
+        </div>
+      </div>
+    </li>
+
+    <li>
+      <div class="group row">
+        <div class="col-1">
+          <p>{{ t.pagecontent.caskcreate }}</p>
+        </div>
+        <div class="col-2">
+{% highlight bash %}
+$ brew cask create foo
+Editing /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/Casks/foo.rb
+{% endhighlight %}
         </div>
       </div>
     </li>
--- a/_data/locales/ar.yml
+++ b/_data/locales/ar.yml
@@ -11,6 +11,8 @@ ar:
     formula: "وصفات \"formulae\" الـHomebrew عبارة عن سكربتات Ruby بسيطة:"
     editor: opens in $EDITOR!
     complement: Homebrew مكمل للـMac، ثبّت الـgems بواسطة <code>gem</code> وثبّت متطلباتها بواسطة <code>brew</code>.
+    caskinstall: '"To install, drag this icon..." no more. <code>brew cask</code> installs macOS apps, fonts and plugins and other non-open source software.'
+    caskcreate: Making a cask is as simple as creating a formula.
     install:
       install: ثبّت Homebrew
       paste: الصق ذلك في الطرفية
--- a/_data/locales/az.yml
+++ b/_data/locales/az.yml
@@ -11,6 +11,8 @@ az:
     formula: "Homebrew formulaları sadə Ruby proqramlarıdır:"
     editor: opens in $EDITOR!
     complement: Homebrew macOS-i tamamlayır. Gemlərinizi <code>gem</code> ilə, onların asılılıqlarını isə <code>brew</code> ilə yükləyin.
+    caskinstall: '"To install, drag this icon..." no more. <code>brew cask</code> installs macOS apps, fonts and plugins and other non-open source software.'
+    caskcreate: Making a cask is as simple as creating a formula.
     install:
       install: Homebrew-i yükləyin
       paste: Bunu Terminal-a əlavə edin.
--- a/_data/locales/be.yml
+++ b/_data/locales/be.yml
@@ -11,6 +11,8 @@ be:
     formula: "Формулы Homebrew — гэта простыя скрыпты на Ruby:"
     editor: адкрыецца ў $EDITOR!
     complement: Homebrew дапаўняе macOS. Усталёўвайце гемы камандай <code>gem</code>, а іх залежнасці — <code>brew</code>.
+    caskinstall: '"To install, drag this icon..." no more. <code>brew cask</code> installs macOS apps, fonts and plugins and other non-open source software.'
+    caskcreate: Making a cask is as simple as creating a formula.
     install:
       install: Усталяванне Homebrew
       paste: Скапіруйце і ўстаўце гэта ў тэрмінал.
--- a/_data/locales/bg.yml
+++ b/_data/locales/bg.yml
@@ -11,6 +11,8 @@ bg:
     formula: "Homebrew формулите са прости Ruby скриптове:"
     editor: opens in $EDITOR!
     complement: Homebrew допълва macOS. Инсталирайте gem пакети, използвайки <code>gem</code>, а техните зависимости чрез <code>brew</code>.
+    caskinstall: '"To install, drag this icon..." no more. <code>brew cask</code> installs macOS apps, fonts and plugins and other non-open source software.'
+    caskcreate: Making a cask is as simple as creating a formula.
     install:
       install: Инсталиране на Homebrew
       paste: Поставете това в терминал.
--- a/_data/locales/ca.yml
+++ b/_data/locales/ca.yml
@@ -11,6 +11,8 @@ ca:
     formula: "Les fórmules de Homebrew són senzills <i>scripts</i> Ruby:"
     editor: opens in $EDITOR!
     complement: Homebrew complementa macOS. Instal·la les teves <i>gems</i> amb <code>gem</code>, i les seves dependències amb <code>brew</code>.
+    caskinstall: '"To install, drag this icon..." no more. <code>brew cask</code> installs macOS apps, fonts and plugins and other non-open source software.'
+    caskcreate: Making a cask is as simple as creating a formula.
     install:
       install: Instal·la Homebrew
       paste: Enganxa el codi a una finestra de Terminal.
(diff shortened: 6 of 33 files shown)


reitermarkus:
> what you think are the hard blockers

https://caskroom.github.io/search needs to be migrated to `formulae.brew.sh`, which would need to be renamed to `packages.brew.sh` or something similar. 

The docs also should be migrated or at least there should be a link to them.

vitorgalvao:
> CC @Homebrew/cask maintainers for thoughts on this

I’m glad to see the page go. I was never a fan.

I like how it looks in your example, but now that I’m taking a closer look, we should really change the text and examples.

1. I’d prefer if we gave something other than `google-chrome` as the example. Let’s make it something a bit more ethical, like `firefox` or `atom`. Maybe even include [the gif](https://github.com/Homebrew/homebrew-cask#lets-try-it).
2. The whole second part is awful. We install a bunch of different crap and should mention that. Also, the `ls` example is helpful to no one. No one should ever need to look inside `/usr/local/Caskroom`.
3. There’s no point to saying users can improve `brew cask` itself, since it’s part of `brew`. We might also want to consider including `cask-repair` in that section about updating casks, since a bunch of users use it frequently.

adidalal:
No real opinion on software showcased. Maybe pick something that's popular, but with a complicated? install.

We should mention: fonts, drivers, etc.

`brew cask list` instead of `ls`?

Should we mention `upgrade`?

+1 for cask-repair mention, even if it's mentioned near the brew equivalent (bump).

MikeMcQuaid:
> https://caskroom.github.io/search needs to be migrated to formulae.brew.sh, which would need to be renamed to packages.brew.sh or something similar.

It's not going to get renamed, I'm afraid. It's going to be a significant amount of work to get all that in place which would need to happen first. Do you have Google Analytics on this site? How widely used is the search? Do all @Homebrew/cask maintainers agree it needs to be a hard blocker on the migration? Would you consider moving this across first and adding the search later if it's desired/requested by users?

> I’d prefer if we gave something other than google-chrome as the example. Let’s make it something a bit more ethical, like firefox or atom. 

Google Chrome is much more widely used; I'd rather we stick to usage rather than subjective "ethicalness".

> Maybe even include the gif.

Would rather keep the style on the page consistent with other examples.

> Also, the ls example is helpful to no one. No one should ever need to look inside /usr/local/Caskroom.

Agreed.

> There’s no point to saying users can improve brew cask itself, since it’s part of brew.

Agreed.

> We might also want to consider including cask-repair in that section about updating casks, since a bunch of users use it frequently.

We don't mention e.g. `brew doctor` so I'd rather we stuck to the happy path here.

> The docs also should be migrated or at least there should be a link to them.

https://github.com/Homebrew/homebrew-cask/issues/48659 was opened for doing this so they are part of https://docs.brew.sh.

vitorgalvao:
> Do all @Homebrew/cask maintainers agree it needs to be a hard blocker on the migration? 

I don’t care about having package search online. As far as I’m concerned, searching for a package is to happen in `brew cask search`. Some users do use it, though, but I don’t think we have numbers on how many.

> Would you consider moving this across first and adding the search later if it's desired/requested by users?

Fine by me.

> Google Chrome is much more widely used; I'd rather we stick to usage rather than subjective "ethicalness".

I don’t think any of our users know what Google Chrome is and not Firefox. The recognition should be about the same. Keep in mind I wrote that (and this) on Google Chrome. I’m not a Firefox user (they lack AppleScript support) so I’m not trying to push some agenda with the change.

There’s another reason why I always thought Google Chrome is a bad example, that I forgot to mention. The fact it’s named `google-chrome` (`vendor-app`). That gives the impression packages will be named like that and you’d expect `mozilla-firefox` and `github-atom`, when that’s not the case for the vast majority of casks.

> We don't mention e.g. `brew doctor` so I'd rather we stuck to the happy path here.

Trust us, `cask-repair` *is* the happy path. For all its ugliness inside (implementation-wise) it’s the tool contributors frequently laud as making contributions easy and (most importantly) *enjoyable*. A user that posts an issue about an outdated cask only does it once — they’re redirected to those instructions and are sending PRs on the next interaction.

I feel pretty confident in saying that little script is responsible for the majority of contributions and for keeping casks up to date. Our most frequent contributors, those who have scripts to automate checking for updates and submitting them, all use `cask-repair` as the final step.

On a tangential note, thank you for starting the work of merging the two pages!

---

> We should mention: fonts, drivers, etc.

Not sure about drivers since that’s more of a technical distinction, but fonts for sure, and maybe quicklook plugins (nice power-user surprise).

> `brew cask list` instead of `ls`?

Maybe tapping the font repo and installing a font, to make it clear it’s a different repo and won’t just work out of the box.

> Should we mention `upgrade`?

Nah, don’t think we need to. By this point, the only people that don’t expect `upgrade` to work are some distracted longtime users that probably think it’ll never happen!

ilovezfs:
Do we mention `bump-formula-pr`? If not I'm not sure why we'd mention `cask-repair`.

ilovezfs:
I'd say the main problem is the name. `cask-repair` really makes it sound like something is broken.

vitorgalvao:
> I'd say the main problem is the name. `cask-repair` really makes it sound like something is broken.

And many times, it is. It doesn’t just bump versions, it also allows you to fix other problems before submitting. It’s power lies in letting you mix and match automated and manual fixes in a single continuous interaction.

reitermarkus:
> searching for a package is to happen in `brew cask search`.

Which FYI, doesn't exist anymore as of yesterday, just use `brew search` now. 😉

ilovezfs:
>And many times, it is. It doesn’t just bump versions, it also allows you to fix other problems before submitting. Its power lies in letting you mix and match automated and manual fixes in a single continuous interaction.

I don't doubt it but the name really disqualifies it for mention here, I think.

claui:
> I don't doubt it but the name really disqualifies it for mention here, I think.

I think I see where you’re coming from, and that you may find the name not suited for presentation.

On the other hand, I feel it’s super important to set expectations what Homebrew is and what it is not. And this includes making people aware that casks may break any moment, or may deliver outdated versions – until someone comes along and repairs it.

Looking brittle > pretending not to be brittle.

ilovezfs:
>I don’t think any of our users know what Google Chrome is and not Firefox. The recognition should be about the same.

It's more about matching what people are likely to want to install as opposed to there being a likelihood of anyone thinking "Firefox, what's that?"


>Looking brittle > pretending not to be brittle.

There are places far more appropriate in the documentation than the main homepage for discussing "repair." Not discussing repair on the main page is not "pretending not to be brittle."

claui:
> There are places far more appropriate in the documentation than the main homepage for discussing "repair."

How about we deep link to it, then?

`<p><a href="https://github.com/Homebrew/homebrew-cask/blob/master/CONTRIBUTING.md#adding-a-cask">Creating new Casks</a> is only a matter of minutes; <a href="https://github.com/Homebrew/homebrew-cask/blob/master/CONTRIBUTING.md#updating-a-cask">updating them</a>, of seconds!</p>`

ilovezfs:
That would be better than mentioning repair, yes.

Still, I think the target audience for the main page is the vast majority of users, who aren't going to be submitting update PRs.

commitay:
I think Firefox would be a better example to use as it will always work, Chrome could have a checksum mismatch when there has been a new release and the cask hasn't been updated yet.

ilovezfs:
Definitely Opera.

vitorgalvao:
> Which FYI, doesn't exist anymore as of yesterday, just use `brew search` now.

You’re moving fast!

> It's more about matching what people are likely to want to install as opposed to there being a likelihood of anyone thinking "Firefox, what's that?"

Are you really arguing that a macOS user on their own who understands Homebrew’s page enough to not be afraid to open a terminal and type some commands to install software can also know enough to know there are alternative (as in, not default with the system) browsers and have heard of Google Chrome and understand enough to know they want it along with other software (otherwise, why bother using `brew cask`), but have never heard of Firefox?

If it can be proven to me without a shadow of a doubt such a person exists (no cheating!) I’ll ask a friend to paint my fingernails bright pink (I’m a bearded short-haired heterosexual man) for a week. I’ll also want to interview said person because (naturally) I stand to learn from a segment of users I don’t believe exists.

Let’s be realistic. If the current Caskroom homepage already said `firefox`, would you be pushing to change it to `google-chrome`? So why not grant that small inconsequential (to you) change?

And if you really think `firefox` is an unknown, pick *any* other cask that doesn’t follow the `vendor-appname` nomenclature and it’s the same to me. Surely the argument isn’t that `google-chrome` is the only well-known software.

ilovezfs:
I think you misread what I said. I specifically said they _will_ know what Firefox is but we should still try to match what most of them want.

vitorgalvao:
> I think you misread what I said. I specifically said they will know what Firefox is but we should still try to match what most of them want.

Had to reread it a few times but yes, I now I understand what I misread.

> I think Firefox would be a better example to use as it will always work, Chrome could have a checksum mismatch when there has been a new release and the cask hasn't been updated yet.

I wasn’t going to make that argument because my post was already long, but yes, Chrome has historically not been the most pristine example of a cask. It has been versioned for less than half of its existence, and even then its versioning is reliant on an appcast that isn’t even hosted on a `google` domain to match an unversioned URL.

To further expand, this means there’s a low chance a user typing `brew cask install firefox` will get an error, while there’s a reasonable chance a user running `brew cask install google-chrome` will get an error. Not a great first experience.

MikeMcQuaid:
Let’s use Firefox.

_layouts/index.html
```diff
@@ -112,6 +119,50 @@ <h2 id="question">{{ t.pagecontent.question }}</h2>
       </div>
     </li>
 
+    <li>
+      <div class="group row">
+        <div class="col-1">
+          <p>“To install, drag this icon…” no more! <code>brew cask</code> brings Homebrew's elegance, simplicity and speed to macOS applications and large binaries alike.</p>
```
vitorgalvao: Apostrophe in `Homebrew’s` should be `’` instead of `'` (the homepage already in place has the correct usage).

_layouts/index.html
```diff
@@ -112,6 +119,50 @@ <h2 id="question">{{ t.pagecontent.question }}</h2>
...
+$ brew cask install google-chrome
+{% endhighlight %}
+        </div>
+      </div>
+    </li>
+
+    <li>
+      <div class="group row">
+        <div class="col-1">
+          <p><code>brew cask</code> does one thing, does it well, and plays nice with others. Apps are moved to your <code>/Applications</code> folder in the same way as you would manually.</p>
```
ilovezfs: Oxford comma is against the style guide! 😱

_layouts/index.html
MikeMcQuaid: This is done by GitHub Pages. I'll remove other "smart quotes" in here.

MikeMcQuaid:
Suggested current wording (and have pushed that to this PR):

screen shot 2018-06-19 at 20 04 18

When we're (mostly) agreed I'll add it to the various translations files (which is what we'll do when doing this properly but makes it harder to iterate on and review).

---

> https://caskroom.github.io/search needs to be migrated to formulae.brew.sh, which would need to be renamed to packages.brew.sh or something similar.

Perhaps a middle ground for this would be that we have:

- https://formulae.brew.sh which is as-is
- https://casks.brew.sh which has just the current search
- https://packages.brew.sh which links to the above two and explains the difference between a formula and a cask (and can be linked from the homepage)
- https://caskroom.github.io redirects to https://brew.sh
- https://caskroom.github.io/search redirects to https://casks.brew.sh

casks.brew.sh can be iterated upon as desired to build up the same functionality as formulae.brew.sh and share some code with it once the various building blocks are in place. I don't think it's reasonable for this (e.g. adding JSON output to `brew cask` and building the same infrastructure for formulae) to be a hard blocker for any migration.

Thoughts?

vitorgalvao:
1. I liked the old ““To install, drag this icon…” no more!”. It’s a fun point of entry and a nod to `brew cask` origins. But it’s OK if you prefer to let it go for message consistency.
2. You forgot to close the parenthesis.
3. I find the message a bit long and overly specific. Maybe something like “`brew cask` complements `brew` by supporting GUI apps, fonts, and plugins. It doesn’t contain them in a single directory, instead aiming to mimic the result of a manual installation.”
4. It’s even fine by me if you want to remove the `brew cask create` example. It doesn’t seem necessary to me. But I also don’t think it’s harmful, I can take it or leave it.

Everything else seems OK to me. Would you say about six months of redirection for the caskroom homepage is about right, and then we kill it? The faster we kill it the better, since [the old domain](https://caskroom.github.io/) was bought by someone else and we have no idea what they plan to do with it (so far it’s just hosting the old code).

MikeMcQuaid:
Thanks for the review everyone! Have added the latest version to the homepage, deprecated and archived https://github.com/Caskroom/caskroom.github.io. It can be deleted in a few months when we've confident the necessary search engines have redirected.

I've not created packages.brew.sh, cask.brew.sh or kept the search page as there was only one voice asking for the latter. I have no objections to these being created as long as they use the Homebrew site styling (which is easy to use with Jekyll Remote Themes now).

simnalamburt:
I'm cool with merging two websites but I really miss the cask search page. 😢 I've even sent a PR to it https://github.com/caskroom/caskroom.github.io/pull/66

I want to implement a new one by myself. Is there any ongoing discussion or working group on this topic?

MikeMcQuaid:
> Thanks for the review everyone! Have added the latest version to the homepage, deprecated and archived https://github.com/Caskroom/caskroom.github.io. It can be deleted in a few months when we've confident the necessary search engines have redirected.

I renamed it and moved it to the Homebrew organisation so the site is no longer present.

> I want to implement a new one by myself. Is there any ongoing discussion or working group on this topic?

There's no current work or discussion but we'd welcome one. In an ideal world it would use the underlying data on formulae.brew.sh for formulae and casks which should soon include cask data too. There's already a JSON API for formulae that should be easy to consume and extend to casks. CC @brianmorton who is working on and discussing this in
https://github.com/Homebrew/brew/pull/4620

PR #243, opened 2018-08-16
Fix Chinese text according to user suggestion

zbeekman:
- Please see https://discourse.brew.sh/t/translate-to-chinese-suggest/2790/2
 - This should be reviewed by someone more linguistically gifted than me, if possible


MikeMcQuaid:
Works for me, thanks @zbeekman!

PR #261, opened 2018-10-19
Fix Serbian and Swedish locales

EricFromCanada:
Fixes issue where selecting either language would fail to load the correct translation.


MikeMcQuaid:
Thanks again @EricFromCanada!

PR #275, opened 2018-11-13
Bump mention of Lion to Mountain Lion

EricFromCanada:
Extra steps are also required for Mountain Lion due to its lack of TLS 1.2 support.

Paired with Homebrew/brew#5238.


MikeMcQuaid:
Thanks again!

PR #284, opened 2018-12-16
Update robots.txt

RDIL:
Makes search engines know they can parse everything


MikeMcQuaid:
Thanks so much for your contribution! Without people like you submitting PRs we couldn't run this project. You rock, @RDIL!

PR #285, opened 2018-12-18
Optimize images

RDIL:
Quality is retained but file size is lowered, allowing for faster page load!


MikeMcQuaid:
@RDIL Nice work again!

PR #293, opened 2019-01-02
Add Norwegian Nynorsk locale

semonje:
This adds Norwegian Nynorsk (`nn`) as a locale for Brew.sh, and moves the existing Norwegian Bokmål (`nb`) to the correct locale (from `no` to `nb`). 

_Some additional info: Spoken Norwegian is represented by two official written languages, Nynorsk and Bokmål. The changes are made according to [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)._

### DIFF
--- a/_data/locales/nb.yml
+++ b/_data/locales/nb.yml
@@ -1,5 +1,5 @@
-"no":
-  locale_name: Norsk
+nb:
+  locale_name: Norsk bokmål
   subtitle: macOS’ manglende pakkebehandler
   pagecontent:
     question: What Does Homebrew Do?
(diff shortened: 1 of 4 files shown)


semonje:
I'm noticing that this PR isn't getting merged 😛 

Is it lacking something, @MikeMcQuaid? 🙂

MikeMcQuaid:
Apologies for the delay @semonje, I was checking that this worked as expected on Transifex (https://www.transifex.com/homebrew/homebrew-homepage/languages/) and it does so merging now!

Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @semonje!

PR #304, opened 2019-01-14
Git ignored CNAME

RDIL:


reitermarkus:
What does the CNAME have to do with it being static?

RDIL:
Well we don’t want the CNAME to change unless it is absolutely needed

MikeMcQuaid:
This file needs to be checked into Git so should not be ignored. Please always provide descriptions for PRs in future.

RDIL:
👍 got it.

PR #308, opened 2019-01-20
Translate new Spanish lines.

mrtnpwn:
I only translated `donate` string and updated `what` string.


reitermarkus:
Thank you, @mrtnpwn!

PR #314, opened 2019-01-26
Add references to Linux

MikeMcQuaid:
Homebrew will support Linux as-of 2.0.0 so update the homepage English translation to reference that. 

CC @sjackman for thoughts


sjackman:
> The missing package manager for macOS (or Linux).

The cross-platform package manager for macOS, Linux, and Windows.
The portable package manager for macOS, Linux, and Windows.
The up-to-date package manager for macOS, Linux, and Windows.

Just some thoughts. What you have is fine too. Thanks for this change.

MikeMcQuaid:
@sjackman My thinking with the current version was that Linux is used enough to be warrant mentioning but it subtly confirms that macOS is the "primary" platform (which I think I fair). I'm hesitant to include "Windows" without "Subsystem for Linux" which ends up becoming a bit wordy (and hard to put in an image)

I like the idea of either up-to-date or portable replacing missing but I'm not sold on the current wording.

sjackman:
Perhaps "Windows 10" instead of "Windows Subsystem for Linux".

The portable and up-to-date package manager for macOS, Linux, and Windows 10.
The up-to-date package manager for macOS (as well as Linux and Windows 10).
The up-to-date package manager for macOS (and Linux and Windows 10).
The up-to-date package manager for macOS (plus Linux and Windows 10).

MikeMcQuaid:
The problem is all these versions are too wordy to fit into e.g. a social card and there's something to be said for being concise. Let's leave it as-is for now given "What you have is fine too" and that Linux now has placement it didn't have 👍

PR #315, opened 2019-01-28
Link to the Linux installation instructions

sjackman:


_data/locales/en.yml
```diff
@@ -18,7 +18,7 @@ en:
     install:
       install: Install Homebrew
       paste: Paste that in a Terminal prompt.
-      what: The script explains what it will do and then pauses before it does it. There are more installation options <a href='https://docs.brew.sh/Installation'>here</a> (required for Linux and Windows 10 Subsystem for Linux).
+      what: The script explains what it will do and then pauses before it does it. Install Homebrew on <a href='https://docs.brew.sh/Installation'>Linux and Windows 10 Subsystem for Linux</a>. Read about other <a href='https://docs.brew.sh/Installation'>installation options</a>.
```
MikeMcQuaid: Please reference the fact the other installation instructions aren't for Linux (it feels implied that they are). Additionally, please update the translations using the existing strings.

MikeMcQuaid: > Does that help?

Yep 👍 

> Do I need to indicate somehow that the translations are stale after this change?

If there's any places you can copy-paste fix in like in https://github.com/Homebrew/brew.sh/pull/314 please do so. Otherwise: let me know and I'll merge as-is.

MikeMcQuaid: @sjackman I meant the untranslated bits but it wouldn't quite work given how things have been shuffled around.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks for merging, Mike!

sjackman:
This PR contained a typo. Sigh. Fix https://github.com/Homebrew/brew.sh/pull/316

PR #337, opened 2019-03-12
Add linuxbrew-256x256.png

sjackman:
This image is used by https://github.com/Homebrew/brew/blob/master/docs/Linuxbrew.md


MikeMcQuaid:
Can we not just use https://github.com/Homebrew/brew.sh/blob/master/assets/img/linuxbrew.png or 
https://github.com/Homebrew/brew.sh/blob/master/assets/img/linuxbrew.svg instead?

sjackman:
There is a `homebrew-256x256.png` image here: https://brew.sh/assets/img/homebrew-256x256.png. This PR was modelled after that.

Yes, I think so. The 256x256 is not much smaller in file size (18 kb) than the 1024x1024 (42 kb). The SVG is only 5 kb. This image is already rendering bizarrely large at https://docs.brew.sh/Linuxbrew. I guess the default is to stretch images to a specified width. Can the default be changed? Otherwise the width of this specific image can be changed.

sjackman:
Does the Markdown rendering engine support the width tag? If not we can use HTML.
```markdown
![Linuxbrew logo](https://brew.sh/assets/img/linuxbrew.png){width=256px}
```

MikeMcQuaid:
> I guess the default is to stretch images to a specified width. Can the default be changed? Otherwise the width of this specific image can be changed.

Yes, let's change the default.



> Does the Markdown rendering engine support the width tag?

Not sure, try it and see!

sjackman:
Closed without merging.

PR #342, opened 2019-03-13
assets/css/styles.scss: decrease image size

sjackman:
Don't make them have a minimum width.


assets/css/styles.scss
```diff
@@ -105,8 +105,8 @@ h3 {
   }
 
   img {
-    min-width: 25em;
-    max-width: 35em;
+    min-width: 6em;
+    max-width: 36em;
```
MikeMcQuaid: Why `36`? May be able to remove `min-width` altogether here; the previous assumption I think was all images were wider than the viewport.

sjackman: 36 em is 6 inches (for a 12 pt font size), which is a good width for a full size image. I'm not wed to any of these numbers. What do you suggest?

MikeMcQuaid: Let's leave it as 35em (that's the viewport max width), remove the `min-width` and see if that's enough?

sjackman: Done.

sjackman: Word of warning, this PR is 💯 cowboy coded and untested.

sjackman: 🤠

MikeMcQuaid: Cool. Mind testing it locally (or asking me to do so) and then we can merge?

sjackman: I have thesis deadlines, so I won't test myself today, could do next week though. If you're able to test, that'd be awesome.

assets/css/styles.scss
```diff
@@ -141,6 +140,9 @@ body {
 
 p {
   margin: 0 0 1em 0;
+  .img {
+    text-align: center;
```
jonchang: Setting `text-align: center` on an `img` won't work since that property only applies to the *contents* of an element. Suggest you set `display: block` (to permit setting the margin) and `margin: 0 auto 0 auto` (which sets the left and right margins be nice and spacious)

MikeMcQuaid:
Thanks @sjackman! Removed image centring as it didn't work here and I'm not sure I agree with universal centring.

sjackman:
It wasn't meant to be universal. You need to add `<p class="img"></p>` to centre an image. Thanks for merging, Mike!

sjackman:
`<p style="text-align: center"></p>` works just as well though.

sjackman:
The sponsors logos are stretched. https://docs.brew.sh/Homebrew-on-Linux#sponsors
This PR should have corrected that, no?

MikeMcQuaid:
@sjackman The page likely needs a rebuild e.g. a push to `master`.

sjackman:
The sponsor logos are no longer stretched. Thanks for the fix.

PR #343, opened 2019-03-13
fix appearance of form elements in Gecko

EricFromCanada:
The problem arises becase Safari defaults to applying `-webkit-appearance: searchfield` to the search field, which causes its font size to shrink from 15px to 11px. This also shrinks the width, which is sized using ems. Increasing the field's `font-size` percentage bumps its font size to 13px after 103%, and after that just increases the width. Firefox, on the other hand, does not automatically apply `-moz-appearance: searchfield` and applies the `font-size: 150%` exactly as specified, so the search field looks like an oversized text field. The language dropdown is similarly affected.

This is fixed in Firefox by applying a browser hack that shrinks the font size for both the search field and language dropdown and corrects the width.

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -298,9 +298,19 @@ button::-moz-focus-inner {
 }
 
 #language {
-  font-size: 104%;
-  font-size-adjust: 0.47;
+  font-size: 16px;
   margin: 1em 0;
+}
+
+#search-bar {
+  width: 17em;
+  margin: 0.1em;
+  padding: 0.1em;
+  font-size: 15px;
+  border: 1px solid #222;
+  border-radius: 4px;
+  text-align: center;
+  -webkit-appearance: none;
 }
 
 #selectable {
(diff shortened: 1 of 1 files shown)


jonchang:
I thought it was _supposed_ to be a really big text box 😅

A Firefox-specific hack doesn't seem like a good solution. [Mozilla's guidance](https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/Advanced_styling_for_HTML_forms#Controlling_the_appearance_of_form_elements) suggests that the only acceptable value for this property is `-x-appearance: none`. Then, we apply CSS styling to get it to look the way we want. What do you think?

EricFromCanada:
I think I've been out of web development for too long. 😬

The search field can have its Webkit-specific style unset with `-webkit-appearance: none` and restyled easily enough, but doing the same on the language menu reverts it on Firefox to the old Windows-style controls. The issue is that Webkit's default form element styling messes with font sizes in a way that Gecko's does not, such that a `<select>` with a computed font size of 13px is noticeably smaller on Webkit compared to Gecko. So, some sort of browser-specific styling is required to get consistent font size. If it's browser hacks you want to avoid, an alternative solution is to use `font-size-adjust` to lower the font size, which is only supported in Firefox so far.

```
#language {
  font-size: 104%;
  font-size-adjust: 0.4;
  margin: 1em 0;
}
```

EricFromCanada:
This post documents WebKit's [interesting behaviour](https://css-tricks.com/dropdown-default-styling/#article-header-id-1) of only allowing `<select>` elements to be 10px, 11px, or 14px. Setting `font-size: 104%;` is the smallest increase that bumps it up to 14px, but since its assigned font size is 16px it will appear slightly larger in other browsers unless they support `font-size-adjust`.

MikeMcQuaid:
Tweaked this and good to go now. Thanks @EricFromCanada!

PR #381, opened 2019-06-20
css: fix ul on mobile

jonchang:
On mobile screens, unordered lists lack the 1em padding at the end that
other paragraphs have. Add this padding for better visual flow.

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -522,7 +522,7 @@ span .amp {
   }
 
   ul {
-    margin: 0;
+    margin-left: 0;
     padding: 0;
     padding-left: 1em;
   }


MikeMcQuaid:
> Add this padding for better visual flow.

What am I missing given this is removing `margin` rather than adding `padding`?

What URL can I see this on?

Thanks!

jonchang:
On mobile Safari, I see this on https://docs.brew.sh/Maintainer-Guidelines#quick-checklist among other places. You can also test this in responsive design mode.

This removes a line because the default margin specification for `ul` is 0 0 1em 0 (1 em margin at the bottom), but the `margin: 0` being removed is within a `screen` width media query. So this is removing an explicit override for narrow screens.

MikeMcQuaid:
This changes more than just the margin at the bottom but also adds a margin on the left too. This should be changed to `margin-left: 0` if the intent is just to add margin at the bottom. I think the lack of margin on the left is superior on smaller screens.

jonchang:
@MikeMcQuaid good catch, I've updated this pull request and will merge when green.

jonchang:
Thanks for the review!

PR #414, opened 2019-10-17
Update ja.yml

imbsky:
"JTF Style Guide for Translators Working into Japanese" don't allow spaces between English words and Japanese. This PR fixes that.


MikeMcQuaid:
Thanks again @imbsky!

PR #436, opened 2019-12-18
Title Case "the missing package manager for macOS" description

issyl0:
- This makes the styles here consistent with Homebrew/brew and brew(1).
- See https://github.com/Homebrew/brew/pull/6826 for further context.


MikeMcQuaid:
Thanks again @issyl0!

PR #441, opened 2019-12-29
Fix accessibility issues

J-Kallunki:
- Select-element must contain label.
- Overflow auto scrolling element must have keyboard access with tabindex="0"

### DIFF
--- a/_layouts/index.html
+++ b/_layouts/index.html
@@ -27,11 +27,9 @@ layout: base
           <p>{{ t.pagecontent.what }}</p>
         </div>
         <div class="col-2">
-          <div class="highlight-wrapper" tabindex="0">
 {% highlight bash %}
 $ brew install wget
 {% endhighlight %}
-          </div>
         </div>
       </div>
     </li>
@@ -135,11 +133,9 @@ end
           <p>{{ t.pagecontent.caskinstall }}</p>
         </div>
         <div class="col-2">
-          <div class="highlight-wrapper" tabindex="0">
 {% highlight bash %}
 $ brew cask install firefox
 {% endhighlight %}
-          </div>
         </div>
       </div>
     </li>
--- a/_layouts/index.html
+++ b/_layouts/index.html
@@ -8,7 +8,9 @@ layout: base
       <div class="group row">
         <h2 id="install">{{ t.pagecontent.install.install }}</h2>
         <br>
-        <pre style='clear:both;text-align:center;margin-bottom:0.9em'><code id='selectable' onclick="selectText(this)">/usr/bin/ruby -e &quot;$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)&quot;</code></pre>
+        <div class="highlight-wrapper" tabindex="0">
+          <pre style='clear:both;text-align:center;margin-bottom:0.9em'><code id='selectable' onclick="selectText(this)">/usr/bin/ruby -e &quot;$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)&quot;</code></pre>
+        </div>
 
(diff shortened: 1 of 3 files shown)


_layouts/index.html
```diff
@@ -27,9 +29,11 @@ <h2 id="question">{{ t.pagecontent.question }}</h2>
           <p>{{ t.pagecontent.what }}</p>
         </div>
         <div class="col-2">
+          <div class="highlight-wrapper" tabindex="0">
```
MikeMcQuaid: These don't need `tabindex` (and I'd argue they are actively not desirable to tab between).

_layouts/index.html
```diff
@@ -8,7 +8,9 @@
       <div class="group row">
         <h2 id="install">{{ t.pagecontent.install.install }}</h2>
         <br>
-        <pre style='clear:both;text-align:center;margin-bottom:0.9em'><code id='selectable' onclick="selectText(this)">/usr/bin/ruby -e &quot;$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)&quot;</code></pre>
+        <div class="highlight-wrapper" tabindex="0">
+          <pre style='clear:both;text-align:center;margin-bottom:0.9em'><code id='selectable' onclick="selectText(this)">/usr/bin/ruby -e &quot;$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)&quot;</code></pre>
```
MikeMcQuaid: Can the `tabindex` be added to the `pre` directly?

_layouts/index.html
J-Kallunki: So how do you scroll them without mouse if overflow is set to auto?

_layouts/index.html
J-Kallunki: Yeah it can be set directly it would work fine. This was implemented here because using the same style definitions as other code pre elements.

MikeMcQuaid: That'd be preferable, thanks!

_layouts/index.html
MikeMcQuaid: I can't see any situation in which you'd need to scroll them or be able to resize the page such that they scroll.

_layouts/index.html
```diff
@@ -40,6 +40,7 @@ <h2 id="question">{{ t.pagecontent.question }}</h2>
           <p>{{ t.pagecontent.how }}</p>
         </div>
         <div class="col-2">
+          <div class="highlight-wrapper" tabindex="0">
```
MikeMcQuaid: Don't need tab index on this or any of the ones below.

assets/css/styles.scss
```diff
@@ -195,16 +195,23 @@ table {
...
+  margin: 0;
 }
 
-pre {
-  margin: 0 0 1em 0;
-  background: $black_30;
-  color: $white;
-  padding: 8px 10px;
-  @include border-radius(0.4em);
+.highlight-wrapper {
```
MikeMcQuaid: Can remove the wrapper entirely.

_layouts/index.html
J-Kallunki: Sorry I don't understand how do you scroll the "overflow-x: auto"?

MikeMcQuaid: I cannot reproduce a situation in which they need to scroll?

J-Kallunki: So is it that `overflow-x: auto` then can be removed? https://github.com/Homebrew/brew.sh/blob/master/assets/css/styles.scss#L207

MikeMcQuaid: I tell a lie: no, it shouldn't be because then the `wget.rb` is trimmed. I still don't think a tab index is needed here.

PR #446, opened 2020-01-06
CI: Lint blog posts with Vale

issyl0:
- Requested in https://github.com/Homebrew/brew/pull/6826#issuecomment-564450052.
- Using the styles in Homebrew/brew, run Vale docs linting on blog posts here.
- I went for the easy approach by using a macOS runner for this as it already has Homebrew installed.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,3 +22,16 @@ jobs:
 
     - name: Run HTML Proofer tests
       run: bundle exec rake test
+
+  vale:
+    runs-on: ubuntu-latest
+    container:
+      image: homebrew/brew:latest
+    steps:
+      - name: Set up Git repository
+        uses: actions/checkout@master
+
+      - name: Run Vale linting on Markdown files
+        run: |
+          brew install vale
+          vale --config=$(brew --repo)/.vale.ini _posts/


.github/workflows/tests.yml
```diff
@@ -22,3 +22,14 @@ jobs:
...
+  vale:
+    runs-on: macos-latest
+    steps:
+    - name: Set up Git repository
+      uses: actions/checkout@master
+
+    - name: Run Vale linting on Markdown files
+      run: |
+        brew install vale
+        vale --config=$(brew --repo)/.vale.ini _posts/
```
issyl0:
```
WARNING: Missing or invalid config file.
```

Wonderful. 🙄

MikeMcQuaid:
Would ideally like to avoid this being two runners if there's a non-terrible way of installing on macOS? Alternatively could run all tests on macOS.

MikeMcQuaid:
> installing on macOS

installing on Linux, sorry!

issyl0:
Looking at [the Homebrew/brew prior art for running on Linux](https://github.com/Homebrew/brew/blob/master/.github/workflows/tests.yml#L16), it's just a copy-paste job but it still feels like more effort than this one line check probably deserves?

I'm happy to switch all of these to macOS.

MikeMcQuaid:
> Looking at [the Homebrew/brew prior art for running on Linux](https://github.com/Homebrew/brew/blob/master/.github/workflows/tests.yml#L16), it's just a copy-paste job but it still feels like more effort than this one line check probably deserves?

Is it available in a non-Homebrew package manager so we can install it that way?

Another option would be to use the Homebrew/brew Docker image?

If neither of the above is possible: I have no strong opinions about macOS vs. the above copy-paste job.

issyl0:
> Is it available in a non-Homebrew package manager so we can install it that way?

It is. But we need to reference the config and styles from Homebrew/brew so we need that repo anyway.

> Another option would be to use the Homebrew/brew Docker image?

I've done this.

issyl0:
```
✔ 0 errors, 0 warnings and 0 suggestions in 15 files.
```

🎉

MikeMcQuaid:
Great work, thanks again @issyl0!

PR #480, opened 2020-04-16
Use relative_url filter to support subpath-hosted sites

jasonkarns:
Urls for assets need to have their baseurl prepended in order to support being hosted at subpaths (ie, the default for GitHub Pages without custom domains). However, the current implementation of manually prepending the site.baseurl within the template is problematic for sub-sites that override their site.logo to reference a fully-qualified absolute URL. (Like [formulae.brew.sh does](https://github.com/Homebrew/formulae.brew.sh/blob/f7236a1d434370a62fc61fdae144b3de78fe4402/_config.yml#L50) within their _config.yml.)

This isn't currently a problem for formulae.brew.sh because it is hosted at the root of its subdomain, so its baseurl is empty. However, when a sub-site is not hosted at the root, the outcome is something like `/foo-repohttps://brew.sh/assets/img/homebrew-256x256.png`

Example output of the main site logo from the nodenv fork of formulae.brew.sh:

`src="/formulaehttps://brew.sh/assets/img/homebrew-256x256.png"`

This PR uses the jekyll filter `relativeurl`, which is purpose-built to do the baseurl prepending. Furthermore, as of jekyll 3.7.0, it intelligently _avoids_ prepending any URL that is already a fully-qualified absolute url.

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -23,9 +23,9 @@ algolia:
     {% seo title=false %}
     {% feed_meta %}
     <meta name="viewport" content="width=device-width">
-    <link rel="icon" type="image/x-icon" href="{{ site.baseurl }}/assets/img/favicon.ico">
-    <link rel="apple-touch-icon" href="{{ site.baseurl }}/assets/img/apple-touch-icon.png">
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Makes sense, thanks again @jasonkarns!

PR #493, opened 2020-05-27
Localization: Tiny adjustment of the German localization

lennartstolz:
In c3ef229ce9ffa5e1d448427ad8424ba16799eb65 the term "Sachen" was replaced with "Zeug". 
Unfortunately, the article "die" still refers to the old term.


MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @lennartstolz!

PR #499, opened 2020-06-05
Update Japanese translation

okuryu:
I'm the first translator of this page. Just recently, I was informed that this translation is outdated, so I reviewed and updated it. I've optimized it to include wording such as Linux support.


MikeMcQuaid:
Thanks so much for your contribution! Without people like you submitting PRs we couldn't run this project. You rock, @okuryu!

PR #507, opened 2020-07-06
Use pipe in install instructions

zanchey:
The current brew install instructions [don't work if you are already running fish](https://github.com/fish-shell/fish-shell/issues/7178) (though, as most brew+fish users have installed fish using Homebrew, it comes up pretty rarely).

Changing the instructions to use a pipe improves portability to non-POSIX shells such as fish, elvish, xonsh, and even tcsh.

However! I understand that there are [potential concerns with regard to trust in the server](https://unix.stackexchange.com/a/339276/78890) with the pipe method, and possibly [issues if the connection dies mid-stream - see non-security concerns](https://sandstorm.io/news/2015-09-24-is-curl-bash-insecure-pgp-verified-install) which might require further thought.

### DIFF
--- a/_layouts/index.html
+++ b/_layouts/index.html
@@ -8,7 +8,7 @@ layout: base
       <div class="group row">
         <h2 id="install">{{ t.pagecontent.install.install }}</h2>
         <br>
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Sorry, we'd rather not alter our instructions.



> Changing the instructions to use a pipe improves portability to non-POSIX shells such as fish, elvish, xonsh, and even tcsh.

If you're using one of these (non-default) shells: you should be able to figure out how to install Homebrew yourself.

PR #509, opened 2020-07-18
Add autofocus attribute to search bar

SeekingMeaning:
This lets us start typing as soon as the page loads, like Google or DuckDuckGo


SeekingMeaning:
Doesn't work :(

PR #516, opened 2020-08-18
Add styles for disabled / deprecated formula names

EricFromCanada:
Prefixes names of formulae that are disabled or deprecated with ⛔️ or ⚠️, respectively.


MikeMcQuaid:
Love it, thanks @EricFromCanada!

PR #518, opened 2020-09-05
Add style for diagrams

claui:
This is a companion PR to https://github.com/Homebrew/brew/pull/8627.

grafik

A few notes on the CSS:

- `background-color: white` is necessary because draw.&#8203;io uses that colour to create negative space (e. g. around labels) so the background needs to match.

- The `filter` and `-webkit-filter` properties are a workaround. They’re needed because browsers don’t allow styling across document boundaries (which an embedded `img` always constitutes).

- The filters are applied *after* setting the background colour, which is convenient because then the background colour still matches the negative space inside the SVG.

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -589,6 +589,16 @@ span .amp {
   }
 }
 
+/* Diagrams */
+img[src$=".drawio.svg"] {
+  background-color: white;
+  margin-bottom: 20px;
+  padding: 5%;
+  width: 90%;
+  filter: invert(80%) sepia(60%);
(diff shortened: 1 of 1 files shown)


claui:
Pinging Homebrew’s designated colour officer, @vitorgalvao.

vitorgalvao:
Thumbs up on the colour. Looks consistent with the rest of the site, and I trust you took care to make it so.

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks @claui!

PR #526, opened 2020-09-29
Add copy to clipboard support for install command

sidharthv96:
Related to https://github.com/Homebrew/formulae.brew.sh/pull/330


assets/css/styles.scss
```diff
@@ -210,6 +210,27 @@ div .highlight {
   margin-top: 0;
 }
 
+.inline {
+  display: inline-flex;
+	@include border-radius(0.4em);
```
MikeMcQuaid, line 215: This indentation seems pretty weird?

_includes/install_command.html
```diff
@@ -0,0 +1,24 @@
...
+<script>
+    function copyToClipboard() {
+        navigator.clipboard.writeText(`{{include.code}}`);
+    }
+
+    if (navigator.clipboard) {
+        document
+            .getElementById("copyButton")
+            .style.removeProperty("display");
+    }
```
MikeMcQuaid, line 23: Sorry to keep changing my mind on this but: I think it'd be fine to use that library you originally proposed for this if it made this easier: how about this code instead:

- is in a seperate `.js` file
- if `navigator.clipboard` exists: looks for all elements with class (not ID) `copyable` 
- writes a clipboard button to the right of those elements
- adds the logic to copy the contents of the current element's text (rather than passing it through and duplicating it in Jekyll)

sidharthv96:
image


The copyable idea can be applied to other installs also...

_includes/install_command.html
sidharthv96: Yeah, I've done something similar.

Don't think we need the other library.

assets/css/styles.scss
sidharthv96: image

Wonder why, looks fine locally.

sidharthv96: Will just reformat it again.

sidharthv96: image

 Haha, tabs vs spaces 😂

sidharthv96:
How should I add a js that can be shared across both sites?

MikeMcQuaid:
> Just add in assets and link it?

Yup!

sidharthv96:
How about in formulae ? Cross link it to the brew.sh js?

assets/css/styles.scss
sidharthv96: Fixed.

sidharthv96:
Applied copyable to suitable commands in homepage.

image

MikeMcQuaid:
> Applied copyable to suitable commands in homepage.

Let's leave it just for the installation line at the top 👍🏻

sidharthv96:
> Let's leave it just for the installation line at the top 👍🏻

How about atleast to `brew install wget` and  `brew cask install firefox` too?

It works perfectly, just wanna enjoy the fruits of the generic code. 😅

sidharthv96:
Any specific reason for not wanting it for the other commands?

MikeMcQuaid (commented):
> Any specific reason for not wanting it for the other commands?

I don't see people copy-pasting them. These are things people read before installation so they don't work (yet).

_layouts/index.html
```diff
@@ -8,8 +8,11 @@
       <div class="group row">
         <h2 id="install">{{ t.pagecontent.install.install }}</h2>
         <br>
-        <pre style='clear:both;text-align:center;margin-bottom:0.9em'><code id='selectable' onclick="selectText(this)">/bin/bash -c &quot;$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)&quot;</code></pre>
```
MikeMcQuaid, line 11: Might be nice to keep the `selectText` bit?

sidharthv96: Done.

_layouts/base.html
```diff
@@ -151,6 +151,7 @@ <h1><a href="{{ site.baseurl }}/">{{ site.title }}</a></h1>
         ));
       };
     </script>
+    <script src="{{ "/assets/js/copy.js" | relative_url }}" async></script>
```
sidharthv96, line 154: @MikeMcQuaid will this load properly in formulae ?

Otherwise I could add the js to the script tag above, and avoid a separate js request.

MikeMcQuaid: > Otherwise I could add the js to the script tag above, and avoid a separate js request.

Yeh, good idea 👍🏻. You'll want to make the JS a function called by something like `<body onload="setupCopyables()">`

sidharthv96: Done!

sidharthv96:
> I don't see people copy-pasting them. These are things people read before installation so they don't work (yet).

Once they run the initial install command, they could come back and try the install right?
It's easier for them to have a one click command, and doesn't hurt to have it anyway :)

SMillerDev:
> Once they run the initial install command, they could come back and try the install right?
> It's easier for them to have a one click command, and doesn't hurt to have it anyway :)

I think it makes it sorta confusing to read, you're reading the docs but it's trying to get you to install wget and firefox. What if I don't even want those? What if I'm on Linux and casks don't work for me? I'd say the cleanest solution is to stick with the parts that people definitely want to copy.

MikeMcQuaid:
Tested this locally, this is looking great! A few needed tweaks:

The "selection on click" now also selects the 📋 and goes "deeper" than it did before:
Screenshot 2020-10-01 at 09 13 18

If you zoom in the alignment for the separate section isn't quite right:
Screenshot 2020-10-01 at 09 11 47

It might be nice to have a hover/click visual difference in the icon (even just in the background, maybe)? Check out what GitHub does:

Screenshot 2020-10-01 at 09 12 30
Screenshot 2020-10-01 at 09 12 32
Screenshot 2020-10-01 at 09 12 36

sidharthv96:
image

On Hover
image


Should the Icon be SVG? Do we have fa icons available ?

MikeMcQuaid:
Hover looks good! No SVG icons available. On click could change back to the non-hover colour to indicate we've taken an action.

sidharthv96:
There is a chrome rendering issue (bug?) here.

image
image

--- 
Firefox looks fine
image
image

sidharthv96:
image

Looks like a chromium bug, it's visible inside VSCode also.

sidharthv96:
image

Will revert back after 1s.

sidharthv96:
Adding 2 SVG icons might look more cleaner than using unicode?

MikeMcQuaid:
> Will revert back after 1s.

Good idea!

> Adding 2 SVG icons might look more cleaner than using unicode?

I think the unicode "feels" more like Homebrew.

sidharthv96:
So I guess we can wrap it up then?

MikeMcQuaid:
Looking great! Almost there! A few tiny things:

Screenshot 2020-10-01 at 16 54 39

Still some very slight alignment issues here (on Safari).

Screenshot 2020-10-01 at 16 54 59
Screenshot 2020-10-01 at 16 54 56

The `selectText` selection area seems to have changed. Not a big deal but might be nice to address.

Relatedly, I wonder whether the `selectText` stuff can go away entirely? Given it won't work if you have JS disabled and the clipboard copy is a much nicer version if you don't. Thoughts?

sidharthv96:
> Still some very slight alignment issues here (on Safari).

Safari was adding extra margin for buttons. Fixed.
image


> The selectText selection area seems to have changed. Not a big deal but might be nice to address.

That's because we are using the highlight instead of a plain code block, so there is a pre with a bottom margin. If I select the inner code, it can be back to normal. `elem = elem.getElementsByTagName("code")[0];`
image

> Relatedly, I wonder whether the selectText stuff can go away entirely? Given it won't work if you have JS disabled and the clipboard copy is a much nicer version if you don't. Thoughts?

Haha, that's why I had removed it at first. I'm fine with either approach, but I think it is a bit distracting to have the copy button copy to clipboard and also select the text, the user might get confused and hit copy again. Also, the select text is broken in chrome for me.

sidharthv96:
The emoji spacing is indeed a chrome issue in macOS.
https://bugs.chromium.org/p/chromium/issues/detail?id=596223

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks so much for your amazing first contribution and patience! Without people like you submitting PRs we couldn't run this project. You rock, @sidharthv96!

sidharthv96:
Thanks @MikeMcQuaid !
Really appreciate the time and effort you put into making this so much better than the initial draft. 😄

MikeMcQuaid:
You're welcome, thanks for your patience with all the rounds of review!

PR #528, opened 2020-10-02
Fix mobile view

### LINKED ISSUE #527: Mobile view broken
image

Working on the fix.

sidharthv96:
Mobile: 
image

Desktop:
image

Fixes #527

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -222,7 +222,6 @@ div .highlight {
 .copyable {
   display: flex;
   justify-content: center;
-  
   .highlight {
     overflow-x: auto;
     margin: 0;
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -220,12 +220,12 @@ div .highlight {
 }
 
 .copyable {
-  display: inline-flex;
+  display: flex;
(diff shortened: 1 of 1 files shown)


sidharthv96:
@MikeMcQuaid Should I remove the select text function or the $ prefix from install?

MikeMcQuaid:
> @MikeMcQuaid Should I remove the select text function or the $ prefix from install?

I removed the select text already before merging the last PR. I don't think the `$` prefix needs removed as you'd already be copying only part of the line if you don't use the JS.

assets/css/styles.scss
```diff
@@ -220,12 +220,12 @@ div .highlight {
 }
 
 .copyable {
-  display: inline-flex;
-  @include border-radius(0.4em);
-
+  display: flex;
+  justify-content: center;
+
```
MikeMcQuaid, line 225:
```suggestion

```

MikeMcQuaid (approved):

sidharthv96:
Cool! I didn't notice that you had removed it.

MikeMcQuaid:
Thanks again for the quick fix @sidharthv96!

PR #536, opened 2020-11-08
Make dollar sign inside `.copyable` segments unselectable

vladimyr:
Push dollar sign inside `::before` pseudoelement so double-clicking on command creates text selection without it.

### Before

image

### After

image


MikeMcQuaid:
Nice work, thanks again @vladimyr!

PR #552, opened 2020-11-25
Install from `HEAD`, not `master`

jonchang:
This seems to work fine and shows no differences between these files:

```
diff <(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh) <(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)
```


MikeMcQuaid:
Good call!

PR #561, opened 2020-12-03
Update ru.yml

KovalevArtem:
Changed the order of words - it sounds more correct!


MikeMcQuaid:
Thanks @KovalevArtem!

PR #562, opened 2020-12-09
Enforce TLS1.2 for install

sidharthv96:
Trigger: [HN discussion](https://news.ycombinator.com/item?id=25356757)
Rust is using these flags in their [install script](https://www.rust-lang.org/tools/install)


MikeMcQuaid:
Passing on this, sorry.

PR #574, opened 2020-12-21
2.7.0: fix some mistakes

Rylan12:
Some fixes for the 2.7.0 blog post.

I also added a reference to `brew livecheck` supporting casks as this is a bigger change since 2.6.0.


MikeMcQuaid:
Thanks @Rylan12 😅

PR #713, opened 2021-09-28
Fixing the Kurdish language code

rawandahmad698:
:)

### DIFF
--- a/_data/locales/kur.yml
+++ b/_data/locales/kur.yml
@@ -1,4 +1,4 @@
-en:
+kur:
   locale_name: کوردی
   subtitle:  macOS (یان Linux) بەڕێوبەری پاکێجە  
   pagecontent:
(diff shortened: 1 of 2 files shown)


maxim-belkin:
Question: English language is abbreviated as `en` -- I wonder if this means that we're using ISO 639-1 and Kurdish language should be `ku`?

rawandahmad698:
> Question: English language is abbreviated as `en` -- I wonder if this means that we're using ISO 639-1 and Kurdish language should be `ku`?

You're correct, but there is two dialects, Kurmanji and Central Kurdish which are often confused, I'm using the ISO 639-2 language code (and to locals in my city, kur best defines Central Kurdish)

maxim-belkin:
Sounds good!

maxim-belkin (approved):

MikeMcQuaid:
Thanks @rawandahmad698!

Issue #721, opened 2021-10-30
zsh: command not found: brew


olegKusov:
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/28058268/139556864-1c9aa77e-b5fb-4b6c-84ba-3a8d20ce3966.png">
I installed homebrew on macOS with this command and it shows that brew not found.

**Please, add additional steps to installation guide on homepage.**

There is "NEXT STEPS shown on command line, it should show something like this

==> Next steps:
- Add Homebrew to your PATH in /Users/$USER/.zprofile:
    echo 'eval $(/opt/homebrew/bin/brew shellenv)' >> /Users/$USER/.zprofile
    eval $(/opt/homebrew/bin/brew shellenv)
    
" (https://stackoverflow.com/questions/36657321/after-installing-homebrew-i-get-zsh-command-not-found-brew). 

**The problem is that users don't read command line. I clear my command line after installation immediately**

MikeMcQuaid:
> The problem is that users don't read command line.

This is a user problem not a Homebrew one.

PR #722, opened 2021-10-30
fix(install): add additinal installation steps

olegKusov:
изображение
Added additional installation steps following #721

### DIFF
--- a/_data/locales/en.yml
+++ b/_data/locales/en.yml
@@ -16,6 +16,7 @@ en:
     caskcreate: Making a cask is as simple as creating a formula.
     install:
       install: Install Homebrew
+      warning: '⚡If shows "zsh: command not found: brew" after trying to install a package with brew:'
       paste: Paste that in a macOS Terminal or Linux shell prompt.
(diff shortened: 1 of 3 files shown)


Bo98:
The problem with putting the commands here is that the one here only applies to people using zsh on arm64 macOS. A different command is needed for Intel macOS and Linux, as well as for other shells that are not `zsh`.

The installer script automatically determines which command is the appropriate one. Maybe it should be clearer to look out for that?

Rylan12:
Alternatively, this could be put in [Common Issue](https://docs.brew.sh/Common-Issues) or the [FAQ](https://docs.brew.sh/FAQ)

olegKusov:
> 
> 
> The problem with putting the commands here is that the one you've put here only applies to people using zsh on arm64 macOS. A different command is needed for Intel macOS and Linux, as well as for other shells that are not `zsh`.
> 
> The installer script automatically determines which command is the appropriate one. Maybe it should be clearer to look out for that?

I understand it. The problem is that people don't look at terminal after installation process is done. Maybe it's better to add some attention label on website or do not finish installation script until user read info about additional steps and then type "yes" .

MikeMcQuaid:
Passing on this, it makes the homepage pretty ugly and I don't see this being a common problem.

PR #864, opened 2022-02-28
s/since 3.4.0/since 3.3.0/

FnControlOption:


MikeMcQuaid:
Oops, good catch, thanks @FnControlOption!

PR #887, opened 2022-07-18
Fix `html-proofer` use in tests

Rylan12:
CI currently [fails](https://github.com/Homebrew/brew.sh/runs/7394015282?check_suite_focus=true) because of issues with `html-parser` [since version 4 was released](https://github.com/gjtorikian/html-proofer/blob/main/UPGRADING.md). I've updated the configuration to be as similar as possible. The main changes: I updated a few options that had been renamed. I removed some options that are now the default or have different functionality. I removed the `check_html` line since the HTML checks [were completely removed](https://github.com/gjtorikian/html-proofer/pull/670) and also removed the `favicon` option since I believe it is covered by including `'Favicon'` in the list of checks.

I needed to set the `allow_missing_href` option since we have several `<link>` attributes without an `href` specified in the translation files. There may be a better way to fix this, but I don't know it off the top of my head.

Finally, I updated a few URLs to use `https`. However, I noticed that the URL http://serkan.io (the homepage for [Serkan Yerşen, one of the translators for Turkish](https://github.com/Homebrew/brew.sh/blob/c68fc5b7c738608b1a94d0d7fb0a2dbc734cc850/_data/locales/tr.yml#L32)) doesn't seem to work for me regardless of whether I use `http` or `https`. I changed it anyway to pass the checks, though.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -13,15 +13,11 @@ task test: :build do
   HTMLProofer.check_directory(
     "./_site",
     parallel: { in_threads: 4 },
-    favicon: true,
-    http_status_ignore: [0, 302, 303, 429, 521],
-    assume_extension: true,
+    ignore_status_codes: [0, 302, 303, 429, 521],
     check_external_hash: true,
-    check_favicon: true,
-    check_opengraph: true,
-    check_html: true,
-    check_img_http: true,
-    url_ignore: [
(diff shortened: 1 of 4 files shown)


mistydemeo (approved):

MikeMcQuaid (approved):
Thanks for this @Rylan12!

PR #893, opened 2022-09-18
Fix search result overflow

kidonng:
Currently, the search result may overflow on small screens:

image

This pull request limits the width of search results, and set a initial scale to prevent similar overflows in the future:

image

To be honest this does not look the best, but it's better than overflowing.

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -22,7 +22,7 @@ algolia:
     {% endif %}
     {% seo title=false %}
     {% feed_meta %}
-    <meta name="viewport" content="width=device-width">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
     <link rel="icon" type="image/x-icon" href="{{ "/assets/img/favicon.ico" | relative_url }}">
     <link rel="apple-touch-icon" href="{{ "/assets/img/apple-touch-icon.png" | relative_url }}">
     <link rel="stylesheet" href="{{ "/assets/css/styles.css" | relative_url }}" type="text/css" media="screen">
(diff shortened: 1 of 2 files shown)


_layouts/base.html
```diff
@@ -22,7 +22,7 @@
     {% endif %}
     {% seo title=false %}
     {% feed_meta %}
-    <meta name="viewport" content="width=device-width">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
```
MikeMcQuaid, line 25: Why is this needed? Can this be scoped just to the list?

assets/css/styles.scss
```diff
@@ -693,4 +693,8 @@ div .algolia-autocomplete {
   }
 }
 
+#algolia-autocomplete-listbox-0 {
```
MikeMcQuaid, line 696: Can we use something related to e.g. `search-bar` instead? This feels like an implementation detail that may change.

MikeMcQuaid (commented):
Thanks for PR! Some thoughts.

assets/css/styles.scss
kidonng: I don't think Algolia is going to change this anytime soon, but I agree we can choose a better selector.

_layouts/base.html
kidonng: This is not needed to fix the bug itself, but to prevent other potential overflows from causing the page to "scale down", like in the first screenshot. Sadly it cannot be scoped.

That said, any overflow needs to be addressed anyway, so it does not matter a lot 😅

MikeMcQuaid: @kidonng If it's not needed to fix this bug: can you remove it? Thanks!

MikeMcQuaid: Oh, you did, sorry!

MikeMcQuaid (approved):
Thanks again @kidonng!

PR #903, opened 2022-11-11
Search: do not set Algolia lang attribute

### LINKED ISSUE #899: The search field doesn't work Spanish
The following pictures explain it. First one, in Spansh, second one in English.

Captura de Pantalla 2022-10-10 a las 16 19 45

Captura de Pantalla 2022-10-10 a las 16 19 58

mistydemeo:
As pointed out by https://github.com/Homebrew/brew.sh/issues/899#issuecomment-1311246326, this attribute prevents us from crawling most of the site when viewed in non-English languages because we don't translate the package database, documentation, etc.

Without this PR, if you set the homepage to any non-English language, you won't be able to search for any package names. With this PR, search should work.

Fixes #899.

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -134,7 +134,7 @@ algolia:
 
       function loadSearch(lang, site) {
         docsearch(Object.assign(
-          { algoliaOptions: { facetFilters: ['lang: ' + lang, 'site: ' + site] } },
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):
great catch, thanks @mistydemeo!

PR #909, opened 2022-12-27
HTML proofer: allow a few known-dead links in old posts

mistydemeo:
As seen in #908, a few of our old links have died since we wrote the old posts they're linked in. That's fine: we can just tell HTML Proofer to ignore them.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -23,6 +23,10 @@ task test: :build do
       "https://hackerone.com/homebrew",
       "https://www.patreon.com/homebrew",
       %r{^https?://twitter\.com/},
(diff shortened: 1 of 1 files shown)


Rylan12 (approved):

MikeMcQuaid:
Could we consider removing the broken links rather than allowing them?

mistydemeo:
I shipped this quickly to unblock the other PR. I planned to do followups today.

I don't think we should remove them - these are in years-old news posts, and it doesn't feel appropriate to substantially alter those old news posts at this point in time. I'm looking into switching them to archive links though.

MikeMcQuaid:
> I'm looking into switching them to archive links though.

Thanks!

PR #910, opened 2023-01-05
Update Kurdish language code to ISO 639-1

samford:
The files related to the Kurdish localization use `kur` as the language code but this should be `ku`. This mismatch leads to a validation error ("Bad value `kur` for attribute `hreflang` on element `link`: The language subtag `kur` is not a valid ISO language part of a language tag.") and this change resolves the issue.


MikeMcQuaid:
Thanks @samford!

PR #911, opened 2023-01-05
Loosen homebrew-science dead link regex

samford:
CI is failing in #910 in relation to `html-proofer` with an `External link https://github.com/Homebrew/homebrew-science failed (status code 404)` error. #909 appeared to be intended to address this but the related regex in the `Rakefile` doesn't match the Homebrew/homebrew-science URL when it doesn't end with a forward slash and trailing text.

This PR modifies the related regex to make the forward slash and trailing text optional, so it will also match the aforementioned URL.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -23,7 +23,7 @@ task test: :build do
       "https://hackerone.com/homebrew",
       "https://www.patreon.com/homebrew",
       %r{^https?://twitter\.com/},
-      %r{https://github.com/Homebrew/homebrew-science/.+},
+      %r{https://github.com/Homebrew/homebrew-science(?:/.+)?},
       %r{https://github.com/Homebrew/brew/.+},
       "https://github.com/mrtnpwn",
       "https://github.com/homebrew/homebrew-php",


mistydemeo (approved):

Rakefile
```diff
@@ -23,7 +23,7 @@ task test: :build do
       "https://hackerone.com/homebrew",
       "https://www.patreon.com/homebrew",
       %r{^https?://twitter\.com/},
-      %r{https://github.com/Homebrew/homebrew-science/.+},
+      %r{https://github.com/Homebrew/homebrew-science(?:/.+)?},
```
MikeMcQuaid, line 26: @samford `%r{https://github.com/Homebrew/homebrew-science}` would be a bit simpler still?

MikeMcQuaid: > The only concern is that it would technically match any Homebrew repository starting with `homebrew-science`,

We don't/won't have any.

> I'm working on a follow-up PR to take another pass at the URLs in the `Rakefile` (e.g., to anchor the start/end of this regex using `^`/`$`, use `(?:/.*)?` here instead (so it would match the URL with only a forward slash at the end), escape dots, etc.).

Is this really necessary? It feels like it's something that's very low impact to our users and fixing a problem that doesn't really exist today?

Note this could well apply to my comment above so: feel free to leave it.

samford: It's not necessary technically (i.e., we only have to account for the existing Homebrew/homebrew-science URLs and not escaping the dots here isn't likely to cause an issue) but it took me less than a minute to make the changes and it may serve as a better pattern to copy if we need to account for similar URLs in the future. If that's not beneficial enough to justify a PR, then there's no harm in leaving it as-is.

MikeMcQuaid: > but it took me less than a minute to make the changes

Fine with me 👍🏻 

> and it may serve as a better pattern to copy if we need to account for similar URLs in the future

I don't think so. I think it's harder to read in a situation where being really strict/precise is unnecessary.

PR #912, opened 2023-01-06
Remove duplicate further-doc id attribute

samford:
The main `index.html` page contains two elements with a `further-doc` `id` attribute. IDs in HTML must be unique, so this leads to a validation error. This commit removes the ID from the element where it seems less appropriate.


MikeMcQuaid (approved):

MikeMcQuaid:
Thanks again @samford!

PR #914, opened 2023-01-09
Replace use of device-width

samford:
The `device-width` CSS media feature is considered deprecated and `width` (e.g., `max-width`) is recommended instead. This PR updates the related `media` query accordingly, resolving a validation error.

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -540,7 +540,7 @@ span .amp {
   padding: 0.1em;
   text-align: center;
   width: 25em;
-  @media only screen and (max-device-width: 480px) {
+  @media only screen and (max-width: 480px) {
     width: 12em;
   }
 }


MikeMcQuaid:
Thanks @samford!

samford:
> Where are you seeing these errors?

I started by simply running the homepage through the [W3C validator](https://validator.w3.org) but then I switched to the command line application (`vnu`), so I could run it across all the HTML/CSS/SVG. The command I've been using is `vnu --skip-non-html --also-check-css --also-check-svg _site` (sometimes with `--errors-only` but that also omits warnings, which are sometimes useful). You can see the error in [the output for `styles.css` before this PR was merged](https://jigsaw.w3.org/css-validator/validator?uri=https%3A%2F%2Fweb.archive.org%2Fweb%2F20230107140313id_%2Fhttps%3A%2F%2Fbrew.sh%2Fassets%2Fcss%2Fstyles.css).

The downside of the CLI app is that they haven't cut a release in the GitHub repository in a couple years (last was 20.6.30 on 2020-06-30) and instead they're regularly updating a "latest" release that's marked as "pre-release". This means that the formula version doesn't always produce the same output as the website. At the moment, it gives an error about usage of `margin-inline-end` (`Property “margin-inline-end” doesn't exist`) though [the validator website doesn't](https://jigsaw.w3.org/css-validator/validator?uri=https%3A%2F%2Fbrew.sh%2Fassets%2Fcss%2Fstyles.css).

I poked around and it looks like they also release the jar on npm (the `vnu-jar` package) and they version those releases, so I've [opened a PR to update the formula](https://github.com/Homebrew/homebrew-core/pull/120151) to obtain the jar from npm. The output from the 22.9.29 version is closer to what the online version produces but they're not identical, since the npm release is three months old (i.e., self-closing void elements are a warning in 22.9.29 but they're "Info" on the website). It's better than the status quo, at least.

I've also been looking around to see if there are any good tools that will go through HTML/CSS/JS to identify any code that may not have great browser support yet. I've been doing this manually by looking through [CanIUse](https://caniuse.com) data but a programmatic approach (with defined browser targets) would be a lot more convenient and could potentially be added to CI. This may also be something to consider (if possible) once I've finished with website tweaks.

PR #915, opened 2023-01-09
Change homepage title text order

samford:
The subtitle comes before the site title in the homepage `title` element, so "Homebrew" comes at the very end (i.e., "The Missing Package Manager for macOS (or Linux) — Homebrew"). This is fine for post titles but "Homebrew" should arguably be the first text of the homepage title, as the full text isn't visible in some common scenarios (e.g., browser tabs, some search results, etc.) and we likely want to prioritize the project name for the homepage.


MikeMcQuaid (approved):
Seems sensible, nice work!

PR #917, opened 2023-01-09
Remove unnecessary script type attribute

samford:
The `type` attribute is unnecessary for `script` elements in HTML5 and using it produces a validation warning. We don't use the `type` attribute with any other `script` elements, so this change brings this element in line with the others.

In practical terms, this is a validation warning that we see across any file using the base template, so it appears in many files and adds a lot of noise to validation checks that aren't restricted to only errors.


Bo98:
Surprised this is a warning tbh since it's not really invalid.

Bo98 (approved):

samford:
> Surprised this is a warning tbh since it's not really invalid.

To my knowledge, invalid markup is treated as errors by the validator and warnings are simply suggestions (i.e., technically valid markup with potential for improvement).

PR #918, opened 2023-01-09
Improve formatting of generated files

samford:
This PR modifies Liquid tags in templates to control whitespace using `-` (e.g., `-%}`), so we don't end up with unnecessary blank lines in generated files. In general, the HTML output should be more readable while also having a smaller file size. There are some technical limitations that prevent us from producing *perfectly* formatted HTML but this is subjectively nicer than the status quo and makes it easier to look through the generated HTML files while debugging.

For what it's worth, this is mostly geared toward improving developer experience while debugging, as all of the unnecessary whitespace from the Liquid tags made the generated files hard to go through. If we wanted to properly reduce the size of files on the production website, we could consider adding a minification step to CI (i.e., so the files are still readable when doing local development but optimized on the deployed site). However, there are other ways to reduce the overall download size of our site (outside of minification) that I'm pursuing first.

-----

~Past that, this modifies the indentation of code blocks in templates so that the generated HTML uses more predictable indentation. In `_layouts/index.html`, this is as simple as indenting the code blocks.~

~However, in `_layouts/base.html`, this involves removing one step of indentation in a couple of `for` loops. This improves the format of the HTML output but the code in the template may be a little harder to read with only one level of indentation. That said, these changes align with how we handle multi-level conditional tags in the `head`.~

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -4,31 +4,31 @@ algolia:
   indexName: brew_all
   inputSelector: '#search-bar'
 ---
-{% assign t = site.data.locales[page.lang][page.lang] %}
+{% assign t = site.data.locales[page.lang][page.lang] -%}
 <!DOCTYPE html>
 <html {% if page.direction == "rtl" %}dir="rtl" {% endif %}lang="{{ page.lang | default: "en" }}">
   <head>
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
-    {% if page.title %}
+    {% if page.title -%}
     <title>{{ page.title }} — {{ site.title }}</title>
-    {% elsif t.subtitle %}
-    {% if page.url == "/" or page.direction == "rtl" %}
+    {% elsif t.subtitle -%}
+    {% if page.url == "/" or page.direction == "rtl" -%}
     <title>{{ site.title }} — {{ t.subtitle }}</title>
-    {% else %}
+    {% else -%}
     <title>{{ t.subtitle }} — {{ site.title }}</title>
-    {% endif %}
-    {% else %}
+    {% endif -%}
+    {% else -%}
     <title>{{ site.title }}</title>
-    {% endif %}
-    {% seo title=false %}
+    {% endif -%}
+    {% seo title=false -%}
     {% feed_meta %}
     <meta name="viewport" content="width=device-width">
     <link rel="icon" type="image/x-icon" href="{{ "/assets/img/favicon.ico" | relative_url }}">
     <link rel="apple-touch-icon" href="{{ "/assets/img/apple-touch-icon.png" | relative_url }}">
     <link rel="stylesheet" href="{{ "/assets/css/style.css" | relative_url }}" type="text/css" media="screen">
-    {% if site.url == "http://localhost:4000" %}
+    {% if site.url == "http://localhost:4000" -%}
     <script src="https://github.com/Khan/tota11y/releases/download/0.1.3/tota11y.min.js"></script>
-    {% endif %}
+    {% endif -%}
     <script>
       (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
       (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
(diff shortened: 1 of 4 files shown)


Bo98 (approved):

Bo98:
Needs a rebase.

samford:
> Needs a rebase.

Rebased but this will need a new approval due to the change.

MikeMcQuaid:
> This improves the format of the HTML output but the code in the template may be a little harder to read with only one level of indentation.

Please optimise for development ease rather than HTML output reading ease, thanks.

samford:
> Please optimise for development ease rather than HTML output reading ease, thanks.

There's a part of me that likes clean generated HTML output but I agree that readability in the template files is more useful (when we have to choose one or the other). I figured we would prefer to improve readability of those Liquid tags, so thanks for confirming.

I've reworked the second commit to adjust whitespace for better readability (and consistency). This includes tweaks to Liquid tag nesting, which should make the structure more understandable.

-----

Besides that, I've also added some general clean-up commits that involve the following:

* Standardizing quotation marks in CSS and JS (if we would prefer to use different quote style for either of these, I can revise this)
* Syntactical tweaks to JavaScript
* Fixing the syntax of a Markdown link in a post
* Removing the trailing forward slash from void elements in our templates (e.g., `<br/>` -> `<br>`), as they're unnecessary in HTML5 and lead to a lot of validation noise. We still have self-closing void elements from the `jekyll-feed` and `jekyll-seo-tag` plugins (where we can't control the output) but this takes care of our own code, at least.
* Removing unnecessary empty attributes (e.g., `class=""`, `value=""`)

I can create a separate PR for these additional changes, if necessary. Otherwise, I figured I could just turn this into a general clean up PR.

MikeMcQuaid:
Sorry @samford but I think this PR has grown larger than it should be without some sort of linter(s) in CI to stop these from regressing. Alternatively, if you don't want to add one now/ever: can you split this into multiple PRs for easier review? Thanks ❤️

samford:
Outside of the previously-mentioned W3C validator, I've been using some HTML/CSS/JS linters to help identify areas to potentially address but I still need to take a deeper look through their configuration options to iron out proper configs. I've done some minimal configuration but I want to make sure that any linters we add are set up to meet our needs. That is to say, I don't want us to run into a situation where our current code passes linting but some code is added/modified that a linter complains about but doesn't actually represent what we want in terms of code style.

I was planning to take care of HTML/CSS/JS changes first, then linter configuration, and integration with CI. It's possible to do the code changes after setting up validation/linting on CI but then any new PRs would be failing on CI until the existing linting issues are resolved, so I figured it would be better to clean up the code before adding more validation/linting to CI.

In the interim time, I've pared back this PR to only the Liquid tag changes to control whitespace in the output. I'll open up separate PRs for the other changes but some of them modify the same code, so I'll need to rebase them as PRs are merged.

MikeMcQuaid (approved):
Looks fine. I'd like to gently ask if this is the most impactful/important work you feel like you could be doing on Homebrew just now but I'm not going to refuse anything on that basis. Thanks!

samford:
> I'd like to gently ask if this is the most impactful/important work you feel like you could be doing on Homebrew just now but I'm not going to refuse anything on that basis.

Long story short, I wanted to take a pass through the website ahead of FOSDEM and I found some areas that can potentially be improved but I figured it wouldn't hurt to clean up the code a bit in the process. I definitely understand the "Isn't there something better you could be doing?" feeling, especially when looking at PRs about whitespace changes, minor tweaks, and cleanup, but I do have some more meaningful changes coming (granted, they may still be seen as minor in the grand scheme of things).

That said, there's some [livecheck] work in Homebrew/brew that I want/need to tackle but sometimes spending time on something a little different can also help to build up momentum for other work. Rest assured, I plan to return to more impactful work after this diversion.

MikeMcQuaid:
> I found some areas that can potentially be improved but I figured it wouldn't hurt to clean up the code a bit in the process

On the flip side:
- this makes `git blame` a fair bit less useful
- all of these changes need reviewed by someone else

> Rest assured, I plan to return to more impactful work after this diversion.

Glad to hear it ❤️

PR #919, opened 2023-01-12
Adjust whitespace for readability/consistency

samford:
This PR contains various whitespace-related tweaks:

* Per our discussion in #918, I've modified the indentation around Liquid tags like `if`/`else`/`for`/etc., so the code can be more easily understood by making the nesting apparent. The generated HTML isn't as nice but we agreed that the readability of the template was more important.
* In the JavaScript, I standardized indentation and added spaces and blank lines where appropriate.
* In `_layouts/index.html`, I modified the formatting for the initial `copyable` `highlight` to use the same style as the other code blocks (for consistency/readability).
* In the post Markdown, I standardized spaces in copy, added a blank line after the front matter, and removed a stray line break.
* In `/assets/css/style.css`, I added blank lines between declarations and selectors and added spaces to a selector with direct descendants (i.e., `li>p>code` -> `li > p > code`).

I can split this into more topical commits, if needed (or if it would help with review).

It's worth noting that this includes changes from #918, so we'll want to merge that first but this should cleanly rebase afterward (or maybe not).

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -10,15 +10,15 @@ algolia:
   <head>
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
     {% if page.title -%}
-    <title>{{ page.title }} — {{ site.title }}</title>
+      <title>{{ page.title }} — {{ site.title }}</title>
     {% elsif t.subtitle -%}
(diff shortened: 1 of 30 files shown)


MikeMcQuaid (approved):
Seems fine, thanks.

Rylan12 (approved):

PR #921, opened 2023-01-12
Remove empty attributes

samford:
This PR adds Liquid tags to conditionally include a `class` attribute on an element in `base.html`, so that we don't end up with `class=""` when `page.header-class` isn't available.

This also removes empty attributes (i.e., `value=""`) from a couple of elements in `newsletter.html`, assuming there's not a technical reason for their inclusion that I'm overlooking.


MikeMcQuaid (dismissed):
Seems fine!

Bo98 (dismissed):

PR #922, opened 2023-01-12
Remove trailing forward slash from void elements

samford:
It's not necessary to self-close a void element in HTML5 and this is the cause of a lot of validation noise for us (unless restricted to only errors). We currently have void elements that use a trailing forward slash and some that don't, so this PR removes the forward slash from related elements.

Unfortunately, our generated HTML will continue to include trailing forward slashes in void elements from the `jekyll-feed` and `jekyll-seo-tag` gems (where we can't control the output) but we can at least address this in our templates.

This is based on #918 and #919 (including changes from both), so it will need to be merged after those PRs but it should cleanly rebase after #919.

### DIFF
--- a/_includes/newsletter.html
+++ b/_includes/newsletter.html
@@ -13,5 +13,5 @@
     <input type="text" name="b_d167156eb854fd2b002d85938_8b2424147f" tabindex="-1">
   </div>
 
-  <input id="newsletter-submit" type="submit" name="subscribe" value="Subscribe" />
+  <input id="newsletter-submit" type="submit" name="subscribe" value="Subscribe">
 </form>
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -9,7 +9,7 @@ algolia:
 <!DOCTYPE html>
 <html {% if page.direction == "rtl" %}dir="rtl" {% endif %}lang="{{ page.lang | default: "en" }}">
   <head>
-    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
     {% if page.title -%}
       <title>{{ page.title }} — {{ site.title }}</title>
     {% elsif t.subtitle -%}
@@ -27,7 +27,7 @@ algolia:
     <link rel="icon" type="image/x-icon" href="{{ "/assets/img/favicon.ico" | relative_url }}">
     <link rel="apple-touch-icon" href="{{ "/assets/img/apple-touch-icon.png" | relative_url }}">
     <link rel="stylesheet" href="{{ "/assets/css/style.css" | relative_url }}" type="text/css" media="screen">
-    <link rel="preconnect" href="https://{{ layout.algolia.appId }}-dsn.algolia.net" crossorigin />
+    <link rel="preconnect" href="https://{{ layout.algolia.appId }}-dsn.algolia.net" crossorigin>
     {% if site.url == "http://localhost:4000" -%}
       <script src="https://cdn.jsdelivr.net/npm/@khanacademy/tota11y/dist/tota11y.min.js" crossorigin="anonymous" async></script>
     {% endif -%}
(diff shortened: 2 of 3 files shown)


MikeMcQuaid:
> Unfortunately, our generated HTML will continue to include trailing forward slashes in void elements from the `jekyll-feed` and `jekyll-seo-tag` gems (where we can't control the output) but we can at least address this in our templates.

Is it really worth addressing here then?

Bo98 (dismissed):
This is fine as is tbh, even if it just means establishing a consistent styling across our templates, instead of being a bit random.

Bo98:
> our generated HTML will continue to include trailing forward slashes in void elements from the `jekyll-feed` and `jekyll-seo-tag` gems (where we can't control the output)

This is because these need to maintain compatibility with XHTML. You'll just have to ignore them.

MikeMcQuaid:
> You'll just have to ignore them.

or submit upstream PR(s) to fix them?

samford:
> > Unfortunately, our generated HTML will continue to include trailing forward slashes in void elements from the `jekyll-feed` and `jekyll-seo-tag` gems (where we can't control the output) but we can at least address this in our templates.
> 
> Is it really worth addressing here then?

I think there's still something to be said for having consistency in our own code and it doesn't hurt to cut down on the validation noise a little bit.

> This is because these need to maintain compatibility with XHTML. You'll just have to ignore them.

> or submit upstream PR(s) to fix them?

I took a quick glance earlier and it seems like it would only require adding a configuration option, modifying the [gem's] template to use the option, and updating related tests. I'm definitely not the first person to want control over this behavior, so I was thinking about creating upstream PRs sometime in the future (unless I run into some kind of blocker in the process).

MikeMcQuaid:
To be honest I'm 👎🏻 on this change at least until it's fixed upstream too, sorry.

Bo98:
Meh, personally I would prioritise template consistency over caring about what libraries do under the hood.

I see this as defining a consistent style. Right now it's 50:50 what style we use.

colindean:
Looking into this a bit, [jekyll-seo-tag's template](https://github.com/jekyll/jekyll-seo-tag/blob/master/lib/template.html) has self-closing tags and these are ~needed because Jekyll _technically_  supports XHTML still (evidenced [here](https://github.com/jekyll/jekyll/blob/9587a73c53d652699d2441458205d4423b888641/lib/jekyll/page.rb#L25-L29) and [here](https://github.com/jekyll/jekyll/blob/6855200ebda6c0e33f487da69e4e02ec3d8286b7/test/test_page.rb#L310-L318), no official documentation alluding it beyond mentions in the changelog.

I opened https://github.com/jekyll/jekyll-seo-tag/issues/486 about it.

colindean:
@samford and I talked about this earlier... he's got a fork of jekyll-seo-tag with the necessary changes already made (and very similar to my napkin plan) so we'll see if the authors respond on the aforementioned issue. If they're receptive, then perhaps Sam can put up the change and we'll go from there. 

Much ado about slashes, haha.

samford:
I'm currently seeing if I can reimplement my Jekyll plugin changes to use parameters (like you suggested in the issue) instead of `site` values (my current implementation), since that's maybe a bit nicer.

colindean:
@samford I know there are a bajillion other far more important things…

and I'm commenting to keep this PR open.

MikeMcQuaid (approved):

MikeMcQuaid:
Once conflicts are resolved: let's just merge this. I don't care enough either way.

MikeMcQuaid:
Thanks for quick fix here @samford 🙇🏻

samford:
Once I've finished working through some `brew` (and homebrew/cask) PRs, I'll see about wrapping up my changes on the `jekyll-feed` and `jekyll-seo-tag` gems to make it possible to control those slashes. It's still an open question whether upstream would merge the changes but it doesn't hurt to try.

Outside of that, I still need to look into adding HTML validation (using `vnu`) to CI.

PR #923, opened 2023-01-12
Update versioned JavaScript libraries

samford:
This PR updates versioned JavaScript libraries in `script` elements (`tota11y`, `anchor-js`) in addition to the following changes:

* Use `cdn.jsdelivr.net` where a specific source isn't required. Minimizing the number of remote sources will cut down on DNS lookups, which doesn't hurt. [If we would prefer to use a different CDN, I can revise this.]
* ~Move `tota11y` from the `head` to the end of the `body` (i.e., I don't see a technical reason why `tota11y` should be loaded in the head).~
* Unversion `tota11y`, since it's only loaded in a local development environment (i.e., any breaking changes won't affect the public site if/when a new version is released).

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -29,7 +29,7 @@ algolia:
     <link rel="stylesheet" href="{{ "/assets/css/style.css" | relative_url }}" type="text/css" media="screen">
     <link rel="preconnect" href="https://{{ layout.algolia.appId }}-dsn.algolia.net" crossorigin />
     {% if site.url == "http://localhost:4000" -%}
-    <script src="https://github.com/Khan/tota11y/releases/download/0.1.3/tota11y.min.js"></script>
+      <script src="https://cdn.jsdelivr.net/npm/@khanacademy/tota11y/dist/tota11y.min.js" crossorigin="anonymous" async></script>
     {% endif -%}
     <script>
       (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
@@ -163,8 +163,8 @@ algolia:
         setupCopyables();
       });
     </script>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/anchor-js/4.2.0/anchor.min.js"
-            integrity="sha256-m1eTvwEHwmsw4+XKF7BshClVJEPwTVycveNl0CS0Mkk="
+    <script src="https://cdn.jsdelivr.net/npm/anchor-js@5.0.0/anchor.min.js"
+            integrity="sha256-aQmOEF2ZD4NM/xt4hthzREIo/2PFkOX/g01WjxEV7Ys="
             crossorigin="anonymous"
             onload="loadAnchors()"
             async></script>


_layouts/base.html
```diff
@@ -159,11 +156,14 @@ <h1><a href="{{ site.baseurl }}/">{{ site.title }}</a></h1>
         }
       }
     </script>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/anchor-js/4.2.0/anchor.min.js"
-            integrity="sha256-m1eTvwEHwmsw4+XKF7BshClVJEPwTVycveNl0CS0Mkk="
+    <script src="https://cdn.jsdelivr.net/npm/anchor-js@4.3.1/anchor.min.js"
+            integrity="sha256-0WMZ9PF4b2hTF66Eglv/9H5Vwk6lnOG4AbmCJxo96WQ="
             crossorigin="anonymous"
             onload="loadAnchors()" async></script>
     <script src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"
```
Rylan12, line 163: Should there be an `integrity` attribute here, too?

Rylan12 (approved):
Tentative 👍 but I can't be certain about the CDN preferences or `tota11y` placement since I don't know a ton about those things

MikeMcQuaid:
> Move tota11y from the head to the end of the body (i.e., I don't see a technical reason why tota11y should be loaded in the head).

It's nicer to have it with the other `<script>` unless there's a performance benefit. Given its localhost only: the performance benefit isn't useful here. 👎🏻 on this change, personally.

MikeMcQuaid:
> * Use `cdn.jsdelivr.net` where a specific source isn't required. Minimizing the number of remote sources will cut down on DNS lookups, which doesn't hurt. [If we would prefer to use a different CDN, I can revise this.]

Seems fine. Trade-off is that it may be less likely to be cached/reused than Cloudflare.

_layouts/base.html
samford: We can only use an `integrity` attribute with files that won't/shouldn't change. In this case, we're simply loading the latest 2.x version of the DocSearch library, so the checksum would change when a new version is released. To use a brew analogy, this is like how we can't use `sha256` with unversioned Casks.

samford:
> It's nicer to have it with the other `<script>` unless there's a performance benefit. Given its localhost only: the performance benefit isn't useful here. 👎🏻 on this change, personally.

My thinking was that `tota11y` is a JavaScript library, so it would be more predictable to look for it with the other libraries at the end of the body. However, `tota11y` is almost 300 KB of JavaScript (uncompressed), so it doesn't hurt to have it load last even if it's only used in a development environment. Granted, I usually comment it out until I need to use it.

The only benefit I can see to keeping it in the `head` is that having it next to the Google Analytics JavaScript makes it easy to temporarily comment out both. That's not very compelling to me (i.e., we should probably add conditional logic so that the Google Analytics code isn't included during local development) but I've removed the commit that moves `tota11y` to the end of the body for the time being.

Was there a reason that you prefer to keep `tota11y` in the head?

-----

> Trade-off is that it may be less likely to be cached/reused than Cloudflare.

Assuming you're referring to client-side caching, cache partitioning in modern browsers means that the library won't be reused even if it was already loaded on a different domain (historically a purported benefit of using a CDN). [This has been the case with Safari since 2013](https://bugs.webkit.org/show_bug.cgi?id=110269) and was also introduced in [Chrome 85](https://developer.chrome.com/en/blog/http-cache-partitioning/) (2020-08) and [Firefox 85](https://developer.mozilla.org/en-US/docs/Web/Privacy/State_Partitioning) (2021-01).

Bo98:
Well you both seem to sort of agree?

> It's nicer to have it with the other `<script>` unless there's a performance benefit

> My thinking was that `tota11y` is a JavaScript library, so it would be more predictable to look for it with the other libraries at the end of the body.

My stance isn't really that dissimilar: consistency - either define all (remote) scripts in the head or all scripts at the end of the body.

samford:
> Well you both seem to sort of agree?

I assumed "the other `<script>`" was referring to the Google Analytics code in the header (i.e., the existing location for the `tota11y` script), so Mike and I may have differing opinions.

> My stance isn't really that dissimilar: consistency - either define all (remote) scripts in the head or all scripts at the end of the body.

I agree with the exception of the Google Analytics code, as they suggest putting it in the `head`. Basically, the code will execute and log the page view sooner with it in the head, whereas we could potentially miss a page view if someone opens the page and navigates away (or closes it) before it's loaded enough for the page view to be logged.

If it's not crucial that we log absolutely every page view (we already miss anyone using an ad/tracking blocker 🤷), then we could consolidate all the script tags. Otherwise, there's something to be said for moving the Google Analytics code to the end of the `head` element, as the current location is a little weird.

colindean (approved):
LGTM 👍

_layouts/base.html
```diff
@@ -160,8 +160,8 @@ <h1><a href="{{ site.baseurl }}/">{{ site.title }}</a></h1>
         }
       }
     </script>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/anchor-js/4.2.0/anchor.min.js"
-            integrity="sha256-m1eTvwEHwmsw4+XKF7BshClVJEPwTVycveNl0CS0Mkk="
+    <script src="https://cdn.jsdelivr.net/npm/anchor-js@4.3.1/anchor.min.js"
+            integrity="sha256-0WMZ9PF4b2hTF66Eglv/9H5Vwk6lnOG4AbmCJxo96WQ="
```
colindean, line 164:
```suggestion
    <script src="https://cdn.jsdelivr.net/npm/anchor-js@5.0.0/anchor.min.js"
            integrity="sha256-aQmOEF2ZD4NM/xt4hthzREIo/2PFkOX/g01WjxEV7Ys="
```

newer version out now

colindean (approved):
New update for anchor.js is available.

I use jsdelivr, it'll route to something else if needed IIRC.

I've used this in the past to judge: https://w3techs.com/technologies/comparison/cd-cdnjs,cd-jsdelivr.

samford:
Updated to use `anchor-js` 5.0.0. I also unversioned the `tota11y` URL (so it will use the latest version), since we don't need to target a specific version and it's only loaded in a local development environment (i.e., any breaking changes when a new version is released won't affect the public website).

MikeMcQuaid (approved):
Seems good. Ideally we'd do this in some way that Dependabot would update but this is an improvement as-is.

MikeMcQuaid:
Thanks @samford!

PR #930, opened 2023-02-06
Use SVG "Fork Me on GitHub" banner

colindean:
This reduces the load size for this image from ~29 KiB to ~7 KiB, but also loads it locally. In theory, this reduces the page load size for https://brew.sh by ~22 KiB, which is around 23% of total size the currently deployed master branch as conveyed through [this GTMetrix report](https://gtmetrix.com/reports/brew.sh/qADTEmpE/).

Image sourced from
<https://tqdev.com/2019-fork-me-on-github-svg-ribbons>, which seems to have an open invitation to use but no explicit license. I've run it through svgo for optimization.

This iteration is slightly different from the Aral Balkan version it replaces in that it's not as thick, but tbh if I didn't have the tabs open next to each other, I'd not have noticed the difference.

New, as rendered in Chrome 109 on my Chromebook: 
Screenshot 2023-02-07 00 18 55

Old: 
Screenshot 2023-02-07 00 18 26


MikeMcQuaid:
Very nice optimisation, thanks @colindean!

PR #931, opened 2023-02-06
Use SVG Homebrew logo for header

colindean:
This reduces the page load size for this element from ~15 KiB to ~4.3 KiB, saving around 10 KiB or reducing the total page load size by around 11% as of the currently deployed master as reported by [this GTMetrix report](https://gtmetrix.com/reports/brew.sh/qADTEmpE/).

The homebrew-256x256.png remains in the repo in case it's been deeplinked somewhere but this header was its only usage in the codebase.

There's no noticeable visual difference between the two, but for me on my Chromebook with Chrome 109 on a high-res screen, it's simply crisper.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -32,7 +32,7 @@ defaults:
     values:
       image: /assets/img/homebrew-social-card.png
 
-logo: /assets/img/homebrew-256x256.png
+logo: /assets/img/homebrew.svg
 
 github:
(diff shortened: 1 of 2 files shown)


MikeMcQuaid (approved):
Makes sense, SVG support is sufficiently widespread. Thanks again @colindean!

PR #934, opened 2023-02-07
Migrate to Jekyll 4 and Actions-based deploy

Bo98:
The GitHub Pages gem is largely legacy and has compatibility constraints - it's stuck on Jekyll v3 indefinitely and that's unlikely to ever change.

The new system to allow GitHub Actions to be used instead for deployment gives a lot more flexibility and allows us to use Jekyll v4 and also not have to stick to whitelisted code (which potentially makes solving issues like https://github.com/Homebrew/formulae.brew.sh/issues/730 and API signing easier).

We already switched to use the Actions deploy on formulae.brew.sh, though still with the pages gem with whitelisting enabled (will open a PR for that soon).

Before merging, we need to change a setting: https://github.com/Homebrew/brew.sh/settings/pages


Gemfile
```diff
@@ -1,8 +1,16 @@
...
+  gem "jekyll-redirect-from"
+  gem "jekyll-remote-theme"
+  gem "jekyll-seo-tag"
+  gem "jekyll-sitemap"
+end
 
 group :test do
   gem "html-proofer"
-  gem "rake"
 end
```
colindean, line 16:
```suggestion
end

group :development do
  gem "rake"
end
```

colindean (approved):
LGTM, perhaps add rake in the development deps just in case.

MikeMcQuaid (changes requested):
What functionality do we actually need from this? I wouldn't agree it's "essentially legacy"; the Actions workflows are still in beta and I'm still not convinced they aren't partly to blame for our intermittent 404 issues.

Bo98:
> What functionality do we actually need from this?

I gave an example:

>> not have to stick to whitelisted code (which potentially makes solving issues like https://github.com/Homebrew/formulae.brew.sh/issues/730 and API signing easier).

I think https://github.com/Homebrew/formulae.brew.sh/issues/730 is a "can't fix" otherwise, or at least not if we remove all JSON templates and produce it all outside of Jekyll. The GitHub pages gem explicitly prevents you from using any Jekyll code not approved by GitHub (understandable in the pre-Actions days).

> it's "essentially legacy";

In terms of the software stack, it is confirmed that there's no plans for it to be updated beyond the usual security/maintenance fixes. Unless the beta doesn't graduate or something, then maybe there will be a revisit.

MikeMcQuaid:
> One would need to take those and run a Jekyll build to get the full end result.

That's what the gitHub-pages one is here, no? https://github.com/Homebrew/formulae.brew.sh/actions/runs/4133603920

Bo98:
Oh oops missed that one since it's uses a different action. Yeah that will work.

Question remains for if we want to add signing, but I'll show on a POC of that first.

colindean:
While I'd like to see this happen, Mike's right: I'm not sure what we gain from moving to Jekyll 4 after reading through [the changelog](https://jekyllrb.com/docs/history/) since the 4.0.0 release in August 2019. Or, rather, that what we'd gain is worth the complexity we'd add by introducing the action. As long as GitHub's keeping up with the Jekyll 3 releases (TIL 3.9.3 came out at the end of January 2023) and the base gems, the best reason we'd have to deviate is if we find a strong need for a Jekyll plugin that GitHub doesn't enable in its default configuration.

Bo98:
The idea was consistency (we already use Actions in formulae.brew.sh and I believe default builds too use Actions now since last year) and to take advantage of the build speed improvements of Jekyll 4 and allow plugins so that we can reduce some of our template boilerplate to single lines (e.g. for formula.json we manually lay out the JSON syntax).

Since then, the direction I think we're taking now is moving API JSON stuff away from Jekyll.

PR #935, opened 2023-02-07
Shave some time by setting up copyables asynchronously

colindean:
GTMetrix reported a main thread block of 106 ms in [this report](https://gtmetrix.com/reports/brew.sh/xEffK6jK/) because of the copyables running synchronously. There's no reason to do that in an age when all browsers we care about support async. So, we catch the DOMContentLoaded and dumbfire the setup.


MikeMcQuaid:
Makes sense thanks @colindean!

PR #937, opened 2023-02-08
Compress images with oxipng, up to 11.62% smaller

colindean:
...but most are 3-6% smaller.

Done with:

    fd '.png' --threads $(nproc) --exec-batch oxipng -o max {}


MikeMcQuaid (approved):
This is great! Would love if there was a bot we could get to do this on any future changes.

PR #940, opened 2023-02-10
Drop Google Analytics

colindean:
It's huge and the stats we get from GitHub Pages are sufficient apparently, per
@MikeMcQuaid.


colindean:
Whoops, misinterpreted something. My mistake.

PR #947, opened 2023-03-11
fix[search box]: Make search box style constant

365cent:
Preview:

[https://brew.b-cdn.net/Screen%20Recording%202023-03-11%20at%206.12.04%20PM.mov](https://brew.b-cdn.net/Screen%20Recording%202023-03-11%20at%206.12.04%20PM.mov)

### DIFF
--- a/assets/css/style.scss
+++ b/assets/css/style.scss
@@ -390,18 +390,51 @@ button::-moz-focus-inner {
   --docsearch-logo-color: #{$white} !important;
   --docsearch-container-background: #{$grey_90} !important;
 }
-
-#search-container button.DocSearch-Button {
-  display: inline-flex;
-  width: 17em;
-  margin: 0;
-  border-radius: 4px;
+#search-container {
+  width: 536px;
+  max-width: 94%;
+  margin: 0 auto;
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
@365cent Can you explain the motivation here? The search box seems to look/work nicely as-is for me?

365cent:
> @365cent Can you explain the motivation here? The search box seems to look/work nicely as-is for me?

In my opinion, the search package interface was more user-friendly when it was just a dropdown list of results. Incorporating Algolia search made it more difficult for me to adapt to, and I suspect that other users may have experienced similar challenges. Making the search box design style constant and placing it in the same location as the open search box button can create a consistent user experience and reduce cognitive load. It can help users to quickly locate the search box and reduce confusion. By forming a mental model of where the search box is located, users can find it easily in the future. This creates a more intuitive and user-friendly interface that can improve the overall user experience. Well, if you believe that the current interface looks good, then perhaps my opinion is not as relevant.

MikeMcQuaid:
> Personally, I prefer the pop-up behaviour of the existing search box.

Agreed, passing on this, sorry @365cent!

PR #949, opened 2023-04-28
Add German translation for search property

WebSnke:
Add German translation for search property.


SMillerDev (approved):

MikeMcQuaid (approved):
Thanks again @WebSnke!

WebSnke:
> Thanks again @WebSnke!

np

PR #957, opened 2023-06-06
layouts/base: move feed tag to include

EricFromCanada:
The `feed_meta` tag throws errors if the jekyll-feed plugin isn't available, but it can be hidden from the parser behind a conditionally-gated include file.


MikeMcQuaid:
@EricFromCanada Woohoo, thanks!

Issue #973, opened 2023-08-20
Implement SRI, says Moz Observatory


colindean:
Quoth Mozilla Observatory, which [gives brew.sh](https://observatory.mozilla.org/analyze/brew.sh) a **D** rating as of [d60d598](https://github.com/Homebrew/brew.sh/commit/d60d598a608541d2e7bca0590f14853f85dc4e90):

Test | Pass | Score | Reason | Info
-- | -- | -- | -- | --
Subresource Integrity | ❌   | -5 | Subresource Integrity (SRI) not implemented, but all external scripts are loaded over HTTPS | Subresource Integrity protects against JavaScript files and stylesheets stored on content delivery networks (CDNs) from being maliciously modified.

<https://infosec.mozilla.org/guidelines/web_security#subresource-integrity> has instructions.

colindean:
It looks like we have _some_ SRI in place already but need to add it one other:

https://github.com/Homebrew/brew.sh/blob/d60d598a608541d2e7bca0590f14853f85dc4e90/_layouts/base.html#L158-L165

colindean:
There's an issue open upstream about this: https://github.com/algolia/docsearch/issues/1561

colindean:
Keep open. I have a plan that might involve moving that in-page js to a file.

MikeMcQuaid:
Don't think we need to keep this issue open given the PR.

PR #974, opened 2023-08-20
Set specific version of Algolia docsearch, set SRI

colindean:
I was about to drop the SRI attributes, but then noticed that jsdelivr had some warnings:

* Skipped minification because the original files appear to be already minified.
* Original file: /npm/@docsearch/js@3.5.2/dist/umd/index.js
* Do NOT use SRI with dynamically generated files! More information: https://www.jsdelivr.com/using-sri-with-dynamic-files

So, to heed these warnings, I decided it was best to pin to a particular version of docsearch and use the pre-minified index.js to avoid dynamic modification.

Work toward #973

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -160,7 +160,9 @@ algolia:
             crossorigin="anonymous"
             onload="loadAnchors()"
             async></script>
-    <script src="https://cdn.jsdelivr.net/npm/@docsearch/js@3/dist/umd/index.min.js"
+    <script src="https://cdn.jsdelivr.net/npm/@docsearch/js@3.5.2/dist/umd/index.js"
+            integrity="sha256-Y1WAhww0aFm/7xcgnD56E3jWSfKlRG9DIB2Tcs8exCQ="
+            crossorigin="anonymous"
             onload="loadSearch('{{ page.lang }}', '{{ page.search_site }}')"
             async></script>
   </body>


colindean:
The one downside to this is getting the security benefits but losing the automatic update of the docsearch library...

And docsearch/js is updated ~monthly: https://www.npmjs.com/package/@docsearch/js?activeTab=versions

Bo98:
The version should be linked to the stylesheet above.

We probably also want a workflow that auto-updates this, with perhaps notices pointing maintainers to test locally in the event of major version bumps.

SMillerDev:
We could add a package.json and fail the build if the versions don't match

MikeMcQuaid:
> We probably also want a workflow that auto-updates this, with perhaps notices pointing maintainers to test locally in the event of major version bumps.

Agreed. It feels like we're going to have to lean into Dependabot if we do this. 

Not convinced the lack of SRI is a bigger security problem than sitting on older versions indefinitely.

colindean:
Upstream docsearch project has an open request for an SRI strategy: https://github.com/algolia/docsearch/issues/1561

colindean:
I posted about a potential solution in https://github.com/algolia/docsearch/issues/1561#issuecomment-1716871326. It's _extra_ but it'd solve the problem if falling behind docsearch versions is a bad thing.

colindean:
If the build-time SRI generation is acceptable, I can move forward with that.

MikeMcQuaid:
> If the build-time SRI generation is acceptable, I can move forward with that.

@colindean Can you detail more what this looks like?

colindean:
See https://github.com/algolia/docsearch/issues/1561#issuecomment-1716871326, but briefly, it's this at `jekyll build` time:

1. retrieve the latest version of the library from the NPM registry API
2. retrieve the content of jsdelivr's cache at that version and get the SRI hash
3. include the output of that in my generated script tag. 

This would safen the resource, creating only a threat vector at site build time.

This also creates a build time dependency on the NPM API and jsdelivr.net, but if these are inaccessible, the build could fall back to using the latest version URL and not specifying an SRI.

MikeMcQuaid:
Closing this out for now but happy to review a new PR when that approach is done.

PR #975, opened 2023-08-20
Adds basic Content-Security-Policy header

colindean:
This should adequately cover the basic CSP use case requiring https for all assets and disallow framing except within our own site.

Possible extensions include limiting loading to just self and cdn.jsdelivr.net with fetch of algolia.net, too. We'd probably also need images from avatars2.githubusercontent.com for blog posts.

Fixes #972

PR #980, opened 2023-08-22
Implement a thorough Content Security Policy in meta

colindean:
OK, this time, I tested it instead of relying on the docs to be clear and accurate like I did for #975 (and suggested what I've ~used in here), necessitating the revert in #979.

I tested it by running `jekyll serve --watch` and using `ngrok HTTP 4000` to get an HTTPS URL to try loading it using TLS. I tested:

1. Algolia search box functionality
2. Clipboard click functionality
3. Language switch functionality (checked with "een beetje Nederlands" and double checked the other items)
4. Paging through many blog posts, but not all

I ensured that there were no console errors while paging around the site.

Mozilla Observatory still dings us for using `unsafe-inline` and there are other CSP-related things we can add, e.g. `base-uri` and `form-action`, which I'll address in future work.

An eventual casualty of this may be the tota11y integration used in local dev, which might be OK because [tota11y is deprecated and EOL since May 2023](https://github.com/Khan/tota11y#deprecation-notice) with a pointer to [Axe](https://www.deque.com/axe/) or to use browsers' built-in a11y tooling which has apparently improved significantly since tota11y came into existence (and @MikeMcQuaid added it six years ago in 3a0c288ffacf208b9d09b6c52fe4fe72a15b3c9f).
I've left it in a working state for now but the workaround is obvious.

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -9,6 +9,24 @@ algolia:
 <!DOCTYPE html>
 <html {% if page.direction == "rtl" %}dir="rtl" {% endif %}lang="{{ page.lang | default: "en" }}">
   <head>
+    <meta http-equiv="Content-Security-Policy"
+          content="default-src 'none';
+                    connect-src https://{{ layout.algolia.appId }}-dsn.algolia.net;
+                    font-src https://fonts.gstatic.com;
+                    img-src 'self' https://avatars2.githubusercontent.com https://avatars.githubusercontent.com;
+                    object-src 'none';
+                    {% comment %}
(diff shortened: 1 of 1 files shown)


colindean:
To be clear, I think this is ready to go, but could use some verification by those more familiar with CSP.

MikeMcQuaid:
> I've left it in a working state for now but the workaround is obvious.

I think it'd be OK to remove at this point but: can be done in a follow-up PR. Thanks again @colindean!

MikeMcQuaid:
Looks good now after merging!

PR #982, opened 2023-08-30
Fixes for anchor-js

EricFromCanada:
1) allow loading fonts via data URLs to the CSP, which is how anchor-js supplies its default link icon; fixes missing hover link icons on docs.brew.sh
2) set the anchor-js visible value to 'hover' since ['touch' was removed for v5](https://github.com/bryanbraun/anchorjs/releases/tag/5.0.0)


MikeMcQuaid (approved):
Thanks again @EricFromCanada!

PR #983, opened 2023-09-06
Add label for language select

### LINKED ISSUE #977: Language selector needs label
[Pagespeed](https://pagespeed.web.dev/analysis/https-brew-sh/rndyyic02f?form_factor=mobile) highlighted that the language `<select>` is missing a label.

samford:
As @colindean mentioned in #977, the language `select` element could benefit from having a label. This PR adds an `aria-label` attribute (for screen readers) as well as a hidden `label` element (for text browsers). The latter is hidden (using CSS) to maintain the current look of the website (assuming we don't want to see a "Language" label next to the `select` box on the page).

This adds a `pagecontent.language` value to the translation files and sets the text to fall back to the English value when it's not present for a given language. I've added translations for all the existing languages, using a combination of Google Translate, [Wiktionary](https://en.wiktionary.org/wiki/language/translations), language-specific dictionaries, websites in the native language (e.g., Wikipedia), etc. Outside of the handful of languages that I have experience with, I can't guarantee that the translations are 100% perfect but I've used everything at my disposal to get as close as possible, at least.

Fixes #977.

### DIFF
--- a/_data/locales/ar.yml
+++ b/_data/locales/ar.yml
@@ -3,6 +3,7 @@ ar:
   subtitle: مدير الحزم الذي يفتقده macOS (أو Linux)
   pagecontent:
     search: بحث
+    language: لغة
     question: ماذا يفعل Homebrew؟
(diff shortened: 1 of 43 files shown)


colindean (approved):
LGTM 👍

MikeMcQuaid (approved):
Thanks @samford!

PR #984, opened 2023-09-06
Replace self-closing tags in plugin output

samford:
The `jekyll-seo-tag` and `jekyll-feed` plugins use self-closing tags (e.g., `<... />`) since they support XHTML but there's no way to override this within the plugins. Self-closing void tags in HTML5 is optional behavior but it creates a bunch of noise in HTML validators unless we restrict the output to errors. Resolving this issue allows us to optionally run an HTML validator in the default mode (which also surfaces warning- and info-level issues) without having to sift through hundreds of lines of output about self-closing tags.

This commit introduces some logic into the `base.html` layout that removes trailing slashes in the plugin output, as this is the only area where we have self-closing tags at this point.

[We will probably want to only check for errors on CI if/when we add it in the future but being able to check for warnings when doing local development is helpful. For what it's worth, I'm running the [W3C validator](https://validator.w3.org) locally using `vnu` (e.g., `vnu --skip-non-html  _site/**/*.html` from within the `brew.sh` directory).]

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -39,10 +39,21 @@ algolia:
     {% else -%}
       <title>{{ site.title }}</title>
     {% endif -%}
-    {% seo title=false -%}
-    {% if site.feed -%}
-      {% include feed.html -%}
-    {% endif %}
+    {% comment -%}
+      The `jekyll-feed` and `jekyll-seo-tag` plugins use self-closing tags (for
+      XHTML compatibility) and, while it's still valid for an HTML5 DOCTYPE,
+      this leads to many `info` notices when running the HTML through a
+      validator. There's no way for us to control this behavior in the plugins,
+      so our only option is to capture the plugin output and naively remove the
+      trailing slash at the end of self-closing tags.
+    {% endcomment -%}
+    {% capture plugin_output -%}
+      {% seo title=false -%}
+      {% if site.feed -%}
+        {% include feed.html -%}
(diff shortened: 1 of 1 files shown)


mistydemeo (commented):
Mind adding a comment why this is being done? I don't think I'll remember why in the future.

samford:
For what it's worth, I'm working on proposing some changes to these plugins that would allow users to opt out of self-closing tags (while still maintaining them as the default for XHTML compatibility) but I'm not yet sure if upstream will be interested in the changes. They haven't commented on a related issue in the `jekyll-seo-tag` repository that's been open for months, so I figured I would create a PR for this workaround in the interim time. We can always replace it if/when the upstream changes are available but, even if they're integrated, there's no telling how long it will take for the plugins to be updated in GitHub Pages as well.

mistydemeo (approved):
Seems like a good workaround in the meantime - thanks for adding the comment.

samford:
Testing this again locally, I just remembered that Markdown images are rendered as a self-closing `img` tag [in Jekyll], so we have to do the same for `content` in the `base.html` template. With that additional change, this finally takes care of all the self-closing tags.

colindean (approved):
Thanks for closing this out. I've had returning to this in my Slack TODOs for far too long.

MikeMcQuaid (approved):
Works for me, thanks @samford!

PR #986, opened 2023-09-06
Rework social card images

samford:
The existing social card images are 1200px x 560px but images that don't use a 1.91:1 aspect ratio can end up being artificially cropped (e.g., https://mastodon.social/@mikemcquaid/110746905748381637). The expected size these days appears to be 800px x 418px or 1200px x 630px.

[I'm not sure why GitHub thinks the SVG file is 571px x 300px but you can check the XML to see that it's using 1200px x 630px dimensions. The text also renders a bit weird but that's to be expected (the file's only intended to be used in vector graphics editors to export a PNG file).]

This PR updates the `homebrew-social-card.svg` file as follows:
* Rework the image to 1200px x 630px and adjust the spacing a bit.
* Replace the embedded Homebrew logo image with the SVG version, which significantly reduces the file size (and makes it resolution independent).
* Switch the text font from SF UI Display to SF Pro Display, as the latter is its updated successor and the former is no longer present on newer versions of macOS, iOS, etc.

This also includes a newly exported PNG file, which has already been run through a compression application.


mistydemeo (approved):
This change makes sense to me.

samford:
Looking into this a bit more, apparently Facebook wants an image that's 1200px x 630px, which is just a few pixels off of a 1.91:1 ratio. I'll take another pass at this in a moment to use that resolution instead (the current 800px x 418px image is very close, so it will only take some small tweaks).

mistydemeo (approved):

colindean (approved):

MikeMcQuaid (approved):
Looks good, thanks again @samford!

PR #988, opened 2023-09-06
Update apple-touch-icon

samford:
The existing `apple-touch-icon.png` file uses a transparent background, so the system automatically creates a background for it. The generated background is black on iOS, so it makes more sense to add a background to the image using the brown color used on the website (#2e2a24).

Besides that, this also repositions the logo to center it in a way that ignores the mug's handle, which makes it _appear_ more centered. The existing image has the image centered based on the entire logo and, while that's technically correct, the logo visually reads like it's not centered.

I created this updated image as a vector graphics file, so I've also added an SVG file.


mistydemeo (approved):

colindean (approved):

MikeMcQuaid (approved):
Thanks again @samford!

PR #990, opened 2023-09-07
Add pretty permalink to alt lang index files

samford:
The existing alternative language files currently use an `/index_{langcode}.html` path. While this is perfectly functional, it's not as pretty as it could be. It's not uncommon for websites to use a language path for alternative languages (e.g., `/fr/`) when they use one domain for all languages.

This commit adds a `permalink` value to the front matter of all the alternative language index files, so they will use cleaner URLs. This also updates the URLs used in the `link` tags and the JavaScript for the language `select` element accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ the index is at `_layouts/index.html`.
 
 If you want to add a new translation, follow these steps:
 
-1. Copy `index.html` to `index_{langcode}.html` and update its `lang` value.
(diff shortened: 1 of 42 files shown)


SMillerDev (approved):

MikeMcQuaid:
Seems fine, thanks @samford. Would be good to get some linting here at some point; you've been adding more and more required attributes for translations and it's unclear what the failure states are if/when any of them are omitted.

PR #993, opened 2023-09-29
Romanian translations updated

vladutilie:
Hello,

I have just made some improvements to the Romanian translations and translated the new strings.

Have a great weekend 🍻


MikeMcQuaid:
Thanks again @vladutilie!

PR #999, opened 2023-11-23
Translated the Portuguese, Portuguese (Brazilian), and Spanish pages

65hz:
# What changes have I made?
I have changed the .pkg notice again but I have made the translations much better and in more languages.

The only Latin language to translate now is Italian, sadly I do not have any Italian experience.


MikeMcQuaid (approved):
Thanks again @65hz!

PR #1000, opened 2023-11-23
Romanian language translations updated

vladutilie:
Hi,

I've made fixes for some typos and updated the translations for the updated strings.

Regards,
Vlad


MikeMcQuaid:
Thanks again @vladutilie!

PR #1007, opened 2024-01-21
favicon: use `svg` logo for favicon when supported

bevanjkay:
Modern browsers allow the use of an `svg` for the site favicon (this is not supported in Safari, so the `ico` will continue to be used).

Before:
Screenshot 2024-01-21 at 9 47 50 pm

After:
Screenshot 2024-01-21 at 9 35 04 pm


MikeMcQuaid (approved):
Good idea, thanks @bevanjkay!

SMillerDev (approved):

mistydemeo (approved):

PR #1008, opened 2024-01-22
Fix broken URLs

MikeMcQuaid:
- Remove some 404ing URLs
- Adjust the `Rakefile` accordingly and to catch new failures


_layouts/index.html
```diff
@@ -217,7 +217,7 @@ <h2 id="analytics-data">{{ t.pagecontent.doc.analytics }}</h2>
       <div class="group row">
         <h2 id="homebrew-donate">{{ t.pagecontent.doc.donate }}</h2>
         <div class="button">
-          <p><a href="https://github.com/homebrew/brew#donations">Homebrew/brew#donations</a></p>
+          <p><a href="https://github.com/Homebrew/brew#donations">Homebrew/brew#donations</a></p>
```
mistydemeo, line 220: Thanks for catching this; I'd mean to do this today. Kind of surprised it didn't follow the redirect!

MikeMcQuaid: @mistydemeo Thanks, you're welcome ❤️

PR #1016, opened 2024-04-11
favicons: update assets & markup

EricFromCanada:
Based on the recommendations in [this exhaustively-researched and recently-updated post](https://dev.to/masakudamatsu/favicon-nightmare-how-to-maintain-sanity-3al7), this does the following:
- updates "favicon.ico" to include 16x16, 32x32, and 48x48 sizes, as generated by [realfavicongenerator.net](https://realfavicongenerator.net/)
- updates "apple-touch-icon.png" with better compression, as generated by previous site
- revises markup to match current recommendations so that most browsers use the SVG and everything else (Safari, IE, others?) use the ICO
Didn't include a "site.webmanifest" as the site isn't intended to be a PWA.

This screenshot of live vs. dev in Firefox for iOS shows a noticeable improvement.

IMG_9296


_layouts/base.html
```diff
@@ -53,10 +53,10 @@
         {% include feed.html -%}
       {% endif -%}
     {% endcapture -%}
-    {{ plugin_output | replace: " />", ">" | replace: "/>", ">" -}}
+    {{ plugin_output | replace: " />", ">" | replace: "/>", ">" }}
```
EricFromCanada, line 56: This change retains the line break preceding the next line in the generated markup.

SMillerDev (approved):

MikeMcQuaid:
Much nicer, thanks again @EricFromCanada!

PR #1030, opened 2024-05-23
Use `.ruby-version` more widely

MikeMcQuaid:
This ensures it's used consistently by `bundle` and ruby/setup-ruby.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,7 +16,6 @@ jobs:
       - name: Install Ruby
         uses: ruby/setup-ruby@943103cae7d3f1bb1e4951d5fcc7928b40e4b742 # v1.177.1
         with:
-          ruby-version: "2.7"
           bundler-cache: true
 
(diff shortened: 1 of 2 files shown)


Bo98:
This is somewhat intentional because we still use the classic method of GitHub Pages deploys here and that use Ruby 2.7.4: https://pages.github.com/versions/

MikeMcQuaid:
@Bo98 GitHub Pages doesn't respect the `Gemfile` at all, though, so I think it's better to have a local environment that works correctly here. We don't have a `Gemfile.lock` here for similar reasons.

I can see an argument for moving to the new method of deploying now it's stable for formulae.brew.sh but I don't think that should block this PR.

Bo98:
It was more to test whether the Pages build will actually succeed, which is after all half the point of this test.

> I can see an argument for moving to the new method of deploying now it's stable for formulae.brew.sh but I don't think that should block this PR.

I agree we should do this.

PR #1126, opened 2025-02-11
add copy command keybind

xplo1t-sec:
Users can now press C on their keyboard to copy the brew install command.

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -202,6 +202,7 @@ algolia:
             const button = document.createElement('button');
             button.innerHTML = escapeHTML('📋');
             button.setAttribute('aria-label', 'Copy to clipboard');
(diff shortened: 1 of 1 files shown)


xplo1t-sec:
We can also add some text to let people know that they can use the C button keybind. Let me know if you have any idea to do this.

bevanjkay:
Thank you for your contribution @xplo1t-sec - but I don't think this change is suitable here.

I'm not a fan of intercepting keypresses on a static website as it is not the expected user experience.
The current implementation will also run the function even if the user is trying to copy something else on the page (using the standard shortcut cmd + c).

xplo1t-sec:
Your point is valid

p-linnane:
Closing per above.

PR #1136, opened 2025-03-13
Improve Italian translation

fpseverino:
- Add references to Linux and Apple Silicon where missing
- Fix grammatical errors and typos
- Make the translation more faithful to the original


p-linnane (approved):
Thanks so much for this @fpseverino! These kinda of contributions are so important and you've actively made Homebrew better!

PR #1206, opened 2025-09-04
Optimise image files

griffenedge:
Optimises image files using [ImageOptim](https://imageoptim.com/mac), making images load faster and saving bandwidth.


MikeMcQuaid (approved):
Thanks!

PR #1279, opened 2026-04-05
Shorten installation URL to brew.sh/install.sh

pragnyanramtha:
This PR simplifies the Homebrew installation command displayed on the website by pointing it to a local wrapper script.

  Changes:
   - Added install.sh: A new, lightweight wrapper script at the repository root. It dynamically
     fetches and executes the official Homebrew install script from GitHub on the fly.
   - Updated _layouts/index.html: Changed the featured installation command from the long GitHub
     URL to the much shorter https://brew.sh/install.sh.

  Why this approach?
   - User Experience: The command is now much shorter and easier to read/share.
   - Reliability: By using a wrapper that fetches the raw script from the official Homebrew/install
     repository, we ensure users always get the most up-to-date and secure version without needing
     manual updates to this repository.
   - Safety: i have retained the /bin/bash -c "$(curl ...)" pattern. This is safer than piping
     directly to bash (curl | bash), as it ensures the script is fully downloaded and parsed before
     execution begins, protecting users from partial execution in case of network failure.

### DIFF
--- a/_layouts/index.html
+++ b/_layouts/index.html
@@ -10,7 +10,7 @@ layout: base
         <br>
         <div class="copyable">
 {% highlight bash -%}
-/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+/bin/bash -c "$(curl -fsSL https://brew.sh/install.sh)"
 {%- endhighlight %}
         </div>
         <br>
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+exec /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
--- a/install.sh
+++ b/install.sh
@@ -1,2 +1,5 @@
 #!/bin/bash
-exec /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+if ! install_script="$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"; then
+  exit 1
+fi
+exec /bin/bash -c "$install_script"


SMillerDev:
> User Experience: The command is now much shorter and easier to read/share.

We haven't had any requests to this extent so I don't really think this is a concern.

> Reliability: By using a wrapper that fetches the raw script from the official Homebrew/install
repository, we ensure users always get the most up-to-date and secure version without needing
manual updates to this repository.

I don't see how that makes it reliable. I do see how this new location can be much easier to compromise for bad actors since there is now a second install script maintainers need to keep track of.

pragnyanramtha:
Hey @SMillerDev, appreciate the review, but i wanted to address something.

On security: The install.sh added here is a thin wrapper that does nothing but fetch and execute the script from the official Homebrew/install repo.

So a compromise of brew.sh/install.sh alone gives an attacker nothing (and would be very hard/suspicious to do), they'd also need to compromise Homebrew/install itself for anything malicious to reach the user. The actual attack surface hasn't changed; this script is just a redirect in shell form.

On user demand: I get that there haven't been formal requests, but the use case is real, Homebrew is increasingly being adopted on Linux, including headless and no-DE environments (servers, WSL, CI setups) where manually typing or copying a long raw.githubusercontent.com URL is genuinely painful. A short, memorable `brew.sh/install.sh` is much more practical in those contexts.

MikeMcQuaid:
> copying a long raw.githubusercontent.com URL is genuinely painful

it's literally one click. passing on this.

pragnyanramtha:
> it's literally one click. passing on this.

I meant it was a pain to copy and manually type it out in distro's without a desktop environment. 

And it looks much cleaner.

SMillerDev:
> So a compromise of brew.sh/install.sh alone gives an attacker nothing (and would be very hard/suspicious to do)

It would give an attacker full control over what runs next. Someone could adjust a gem to rewrite this file and potentially compromise everyone who installs homebrew.

pragnyanramtha:
Understood, thanks for reviewing!

PR #1343, opened 2026-08-01
Gemfile.lock: align with Homebrew/brew

p-linnane:
Matching Homebrew/brew.


ZhongRuoyu:
@p-linnane I think the stack's order may need to be reversed since the one closer to main is the one to be merged first.

Otherwise we can also "unstack" this and merge the Vale fix into this PR.

ZhongRuoyu (approved):

MikeMcQuaid (approved):

PR #1347, opened 2026-08-03
Make 'latest post' of the blog translatable

SMillerDev:
Something I noticed when checking the Ukrainian translations was that this was untranslated. This allows us to translate it.


p-linnane (approved):
Nice!

PR #1359, opened 2026-08-31
styles: add `bottles` CSS id

botantony:
CSS id for `Bottle` (formulae) / `Supported platforms` (casks) table that adds a separator between `<tbody>` tags

See https://github.com/Homebrew/formulae.brew.sh/pull/2282#issuecomment-5476123371


MikeMcQuaid (approved):
Thanks!

PR #1371, opened 2026-09-16
index.html: rename donate id

samford:
The `id` for the heading in the donate section of the homepage is `homebrew-donate`, which feels a bit redundant in this context. It may have simply carried over the naming convention from the previous "Homebrew Packages" and "Homebrew Bundle" sections, where it makes more sense. We may want to link to this URL in the future, so this simplifies the ID.

I did a quick search and didn't see the URL used anywhere on brew.sh but let me know if there's anything else we need to update in relation to this change.


MikeMcQuaid (approved):
Thanks!

PR #1372, opened 2026-09-16
Create standalone donate page

samford:
The brew.sh website currently only contains donation information in a section on the homepage but we discussed creating a standalone /donate page for this as well. This adds the page, sharing an include for the content between the homepage and donate page. The `donate.html` file is very minimal, as we need to add additional classes to the `div` on the standalone page to style it properly (so it can't be part of the include).

I can rework this if we would prefer to hardcode the contents of `_includes/donate.html` in `donate.md` instead of having a separate include.

Also, if anyone has any feedback on the page description, please do let me know. The current description is kind of a generic placeholder and there's probably room for improvement.


pages/donate.md
```diff
@@ -0,0 +1,13 @@
+---
+layout: base
+title: Donate to Homebrew
+description: Support Homebrew, a non-profit project run entirely by volunteers.
+permalink: /donate/
+---
+
+<article id="post">
+  <h2>Donate to Homebrew</h2>
```
MikeMcQuaid:
```suggestion
  <h2>💸 Donate</h2>
```

MikeMcQuaid (approved):
One tweak and good to merge!

pages/donate.md
samford: Shall I update the page title to "Donate" as well (albeit, without the emoji in the title) or leave it as "Donate to Homebrew"? It's worth noting that page titles have a "Homebrew:" prefix, so "Homebrew: Donate" would probably read better than "Homebrew: Donate to Homebrew".

MikeMcQuaid: I think leave page title as-is; "Donate to Homebrew" makes sense to me in the page title. `Homebrew:` prefix is 🤷🏻

1chooo (commented):
Nice and straightforward change. Looks clean to me!

p-linnane (approved):

PR #1377, opened 2026-09-19
Translate German website content

michi-onl:
## Summary
- Translate the 37 remaining English strings in the German locale file
- Drop 7 keys left over from the site redesign
- Keep the informal "du" register already used for `search` and `language`

## Notes
- 37 strings translated: hero, packages, casks, taps, bundle, install, docs, footer
- Stale keys removed: `prefix`, `hack`, `formula`, `editor`, `complement`, `caskinstall`, `caskcreate`. The current `_layouts/index.html` does not render any of them.
- Technical terms stay English: formulae, casks, taps, keg, Cellar, prefix
- HTML, URLs and the `#optional-stanzas` link are unchanged
- The German translation was drafted with assistance from DeepSeek v4.1 Flash


MikeMcQuaid (approved):
Thanks!

